PR #12, opened 2016-12-21
Fixing pagination query performance issues

thebigmoosey:
This fixes an issue in the pagination query that was previously using wildcards when it should be selecting on [Id] and using a subquery to improve performance (this was previously fixed in 3.4, but when we migrated to Nevermore this must have gotten lost in the merge).


thebigmoosey:
Investigated this further and found this could be fixed in a simpler way by improving our indexing in Octopus, closing this PR now.

PR #13, opened 2017-01-13
Do not override values specified in connection string

tothegills:
Relates to OctopusDeploy/Issues#3035


tothegills:
[Override connection string parameters only when they are not provided](https://trello.com/c/1XELzWL2/2-override-connection-string-parameters-only-when-they-are-not-provided)

PR #18, opened 2017-03-02
Raised level of `n transactions active` message and also logged trans…

droyad:
…action dump once to Error

### DIFF
--- a/source/Nevermore/RelationalTransactionRegistry.cs
+++ b/source/Nevermore/RelationalTransactionRegistry.cs
@@ -12,6 +12,7 @@ namespace Nevermore
         readonly ILog log = LogProvider.For<RelationalTransactionRegistry>();
 
         readonly List<RelationalTransaction> transactions = new List<RelationalTransaction>();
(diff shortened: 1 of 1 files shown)


source/Nevermore/RelationalTransactionRegistry.cs
```diff
@@ -41,13 +43,26 @@ public void Remove(RelationalTransaction trn)
...
+        void LogHighNumberOfTransactions(bool reachedMax)
+        {
+            if (reachedMax && !highNumberAlreadyLoggedAtError)
+            {
+                highNumberAlreadyLoggedAtError = true;
+                log.Error(BuildHighNumberOfTransactionsMessage());
+                return;
+            }
+
+            if (!log.IsDebugEnabled())
```
michaelnoonan: Should this be `if (log.IsDebugEnabled())` ?

droyad: Ah yeh. Change isn't tested.

PR #19, opened 2017-03-03
Expose pipe delimiting logic so that it can be used elsewhere (e.g. w…

zentron:
…hen trying to access the raw columns in db update scripts).

Also didnt seem like the HashSetReaderWriter is really needed since its covered by the existing one for referencecollections

### DIFF
--- a/source/Nevermore/Mapping/ReferenceCollectionReaderWriter.cs
+++ b/source/Nevermore/Mapping/ReferenceCollectionReaderWriter.cs
@@ -1,4 +1,5 @@
 
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Nevermore.Mapping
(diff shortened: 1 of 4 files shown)


source/Nevermore/Mapping/ReferenceCollectionReaderWriter.cs
```diff
@@ -15,20 +16,28 @@ public override object Read(object target)
...
             if (collection == null)
             {
                 base.Write(target, collection = new ReferenceCollection());
             }
 
-            collection.ReplaceAll(valueAsString.Where(item => !string.IsNullOrWhiteSpace(item)));
+            collection.ReplaceAll(Parse(value?.ToString()));
+        }
+
+        public static IEnumerable<string> Parse(string value)
```
droyad: What about putting these into a class under `Util` named `CollectionToPipeConverter` or something?

zentron: if you'd prefer. I like having it close to where its used unless it makes sense to seperate rather than for the sake of separation.

PR #21, opened 2017-03-29
Use a static ILog instance for perf

michaelnoonan:
Profiling shows a high JIT cost to creating a new typed-ILog instance per dependency. The logging framework itself is thread-safe so we should just use a static typed-ILog instance.


michaelnoonan:
I thought I already covered all 3 places? :)

michaelnoonan:
Two of them turn up in the hot call chain

PR #22, opened 2017-03-30
Converting projects to new msbuild

johnsimons:
Had a go at doing the conversion, not sure what I missed!

@droyad can you review this ?


droyad:
I'd expect changes to the sln file, the tests and build. Did you have a look at this PR?https://github.com/OctopusDeploy/Halibut/pull/31

johnsimons:
I will @droyad, thanks

johnsimons:
@droyad @slewis74 @alfhenrik this should now be complete.
I have upgraded the cake build script.
I have updated some dependencies.

The test projects need to be net452 now because for xunit2 that is the minimum.

johnsimons:
For some reason the build server doesn't seem to have any compatible agent!

matt-richardson:
@johnsimons I changed the agent requirements to `agentType = Build-VS2017`, and its now kicked building.

johnsimons:
Thanks @matt-richardson 

So it looks like TC is looking in `C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Sdks\Microsoft.NET.Sdk\Sdk\` instead of `C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0`

johnsimons:
This is the issue https://developercommunity.visualstudio.com/content/problem/25594/visual-studio-build-tools-2017-is-missing-msbuild.html

droyad:
Sorry, just saw this convo. We should be using the `DotNetCore` tooling to build these projects, then VS 2017 isn't needed, same as Halibut. Pushed up the changes in the build file, and it looks good now.

johnsimons:
@droyad  Any idea why NUnit report is still being used - https://build.octopushq.com/viewLog.html?buildId=189335&buildTypeId=OctopusDeploy_LIbraries_Nevermore&tab=buildLog#_focus=104 ?

johnsimons:
@droyad regarding using `DotNetCore` I guess we have to for now, but in the future MS is getting rid of it, msbuild for everything, so we need to fix our agents

droyad:
I don't believe that is the case, link? `dotnet` is here to stay.

johnsimons:
That is not my understanding hence the move to csproj vs project.json,  and msbuild.exe vs dotnet.exe and even ms converting msbuild to multi plat. anyway I think for now we are safe

johnsimons:
So do we disable the nunit test reporter in TC feature ?

PR #26, opened 2017-06-23
Normalize spaces in parameter names

tothegills:
IN queries will add parameter names with spaces  eg `@foo 1` which will break the query.  This normalizes spaces in parameter names.


tothegills:
I think strip as we come across them, performance is a big issue.

Issue #34, opened 2018-01-02
Where IN empty list generates invalid Sql


benPearce1:
```
var projectIds = new List<string>();
projects = transaction.Query<Project>()
                                      .Where("Id", SqlOperand.In, projectIds);
```

Results in the following error:
```
{
  "ErrorMessage": "Exception occured while executing a reader for `SELECT * FROM dbo.[Project] WHERE ([Id] IN ()) ORDER BY [Id]`\r\nSQL Error 102 - Incorrect syntax near ')'."
}
```

PR #35, opened 2018-01-02
Better support for WHERE IN against an empty list

benPearce1:
Fixes https://github.com/OctopusDeploy/Nevermore/issues/34


benPearce1:
The Cake.FileHelpers were not being used in the script, but were referenced in the header without specifying a version number. The latest update was not compatible with the Nevermore target framework (i think it is net451 or net45) and specifying the version wasn't working. Script didn't require it anyway.

PR #36, opened 2018-02-19
Allow 1st class subtyping

zentron:


source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
```diff
@@ -83,7 +83,7 @@ RelationalStore BuildRelationalStore(string connectionString, double chaosFactor
                 ? (ISqlCommandFactory)new ChaosSqlCommandFactory(new SqlCommandFactory(), chaosFactor)
                 : new SqlCommandFactory();
 
-            return new RelationalStore(connectionString ?? TestDatabaseConnectionString, TestDatabaseName, sqlCommandFactory, Mappings, new JsonSerializerSettings(), new EmptyRelatedDocumentStore());
+            return new RelationalStore(connectionString ?? TestDatabaseConnectionString, TestDatabaseName, sqlCommandFactory, Mappings, new JsonSerializerSettings(){ TypeNameHandling = TypeNameHandling.Auto }, new EmptyRelatedDocumentStore());
```
zentron: More closely mimics how we use it in Octopus (wanted to make sure that the `$type` property no longer appears in the `JSON` field)

PR #39, opened 2018-02-28
Build queries using an AST and extend query building features

TomPeters:
# Goals
There are 3 main goals I am trying to address with this change:

1. We want to be able to build queries for creating Stored Procedures, Functions, and Views
2. We want to be able to use some more of the features in SQL server
3. We want to make it a bit easier to extend the behaviour of query generation so that we can inject some application-specific business logic into the process

## AST
Extending our query generation code proved to be very difficult in its current form. This PR replaces a lot of the query generation code with an AST which can be used to perform the query generation. The idea is that this new structure is much easier to extend. 

For example, explicit table/subquery aliasing was very difficult to add to the old query generation code. It was much easier to add now that there is an AST to back our query.

All parts of the AST are in the `AST` namespace. The AST is immutable. 

## `IQueryBuilder`
Unfortunately, the `IQueryBuilder` is mutable, and usages of it in other projects rely on the fact that it is mutable, so this is unchanged.

The old model was: `IQueryBuilder` deletages to `SqlQueryGenerator`. `SqlQueryGenerator` was stateful and built up string expressions of parts of the query as it went. At the end, it would plug the main parts of the query together when you are ready to perform a query. All state was stored in `SqlQueryGenerator`.

In the new model we still have an `IQueryBuilder`, which delegates to an `ISelectBuilder` (there are multiple implementations). `ISelectBuilder` stores all state required to build an `ISelect`, which is a root node of the AST that represents a single "select" sql statement. Not all state is stored in the `ISelectBuilder`: Parameters are stored directly on the `IQueryBuilder`, since parameter value are not required in order to build an `ISelect`, but are only required at execution time.

There is some more type safety around what operations you can perform on the `IQueryBuilder`, implemented by sub-types of `IQueryBuilder`. For example, the `On` method can only be called on the `IJoinSourceQueryBuilder`, and you can only get access to the `IJoinSourceQueryBuilder` by calling `Join` on an `IQueryBuilder`. This keeps the typing a bit stricter and prevents needing to handle more edge cases in the code. It also forces the usage to be clearer and less ambiguous (eg the `On` keyword must apply to the specific `Join` that was _just_ added to the query).

## Deletes
In the old code, you could call `.Delete()` at the end of any query. There were a number of things you could do during query generation that meant that the delete query would be invalid, or throw an exception. I have replaced this with a separate query builder just for deletes, which is a stripped down version of the normal query builder. This restricts what you can do with the delete query builder to only valid operations, which will make it much easier to achieve goal 3 ☝️.

## Stored Procs / Functions
Previously, we just had parameters for queries. They had names and values. Nice and simple.

Now it is more complex: If you are creating a stored proc, then you still have parameters, but they don't have values, they only have a name. But they also now must come with a data type as well in this case, and they can also come with a default value. This is all now supported, but one consequence is that I have rename `CommandParameters` (which represented the parameter values only) to `CommandParameterValues` in order to differentiate them from the parameters without values (`Parameters`) and their default values (`ParameterDefaults`).

## Table aliasing
We rely more now on generating aliases for subqueries (this seemed fairly limited before). Previously this was done by producing a random string, but this breaks caching in sql server, so the process is now deterministic for the same query.

## Breaking changes

- Some operations (like joining) create a new `IQueryBuilder` and `ISelectBuilder` and therefore do not continue to mutate the original query builder. This is mostly isolated to things that change the "source" of a query (like joins/subqueries). There are only a limited number of usages in our code of joins that will need to be manually checked and possibly changed.
- Type names changed / new types
    - `IQueryBuilder` has more sub-types, but most code should be unaffected, and any code that is simply chaining queries will be unaffected
    - `CommandParameters` renamed to `CommandParameterValues`
- `Delete` removed from `IQueryBuilder`
- Generated SQL is formatted differently
- Generated SQL may actually be different for more complex queries, but should produce the same result
- Some edge cases for nested queries/joins handled a little differently (and in a more predictable way now IMO), for example see `ShouldGenerateSelectForComplicatedSubqueryJoin`, where the explicit `.OrderBy("Id")` call was applied to the end of the whole query, instead of against the subquery against which it was specified.
- Some ways in which you could invoke the query builder are now not possible. For example
`transaction.Query<MyDocument>().Where("foo = 1").View("SomeView")` will no longer compile, and needs to be replaced with `transaction.Query<MyDocument>().View("SomeView").Where("foo = 1")`
- Order By column names are always escaped in sql. Previously you could do `OrderBy("[Id]")` but now this is invalid, because it will be double escaped in sql (this same code will now end up as `ORDER BY [[Id]]`)
- The new public method on `IRelationalTransaction` to initiate a query is `.TableQuery<TDocument>()`. This allows you to edit the `FROM {table}` part of the query. Most existing places that are doing `transaction.Query<TDocument>()` will now go through an extension method. This extension method uses a tiny hack to force the query builder to be stateful (with respect to adding `Where` statements etc) from the very beginning. This was the only reliable way I could come up with to make sure I did not break anything. The places that use the `.Query<TDocument>()` extension method can't change the "table" part of the query, and those few places need to be updated to use `.TableQuery<TDocument>()`.
- Table alias generator is shared across a single transaction. So multiple queries in the same transaction could affect each other in terms of the aliases generated.


TomPeters:
[Update stored procs/functions/views](https://trello.com/c/9VpKPCrq/35-update-stored-procs-functions-views)

source/Nevermore/AST/Select.cs
```diff
@@ -0,0 +1,26 @@
...
+        public Select(IRowSelection rowSelection, ISelectColumns columns, ISelectSource from, Where where, OrderBy orderBy)
+        {
+            this.rowSelection = rowSelection;
+            this.columns = columns;
+            this.from = from;
+            this.where = where;
+            this.orderBy = orderBy;
+        }
+
+        public string GenerateSql()
```
droyad: It would be nice if `ToString` was overridden to point at `GenerateSql`, that way you can see the query easily in the debugger. `DebuggerView` could also be used to show the main properties of each type, again to help debugging when viewing the AST.

source/Nevermore/SelectBuilder.cs
```diff
@@ -0,0 +1,232 @@
...
+        protected override ISelectColumns DefaultSelect => new SelectAllFrom(From.Source.Alias);
+
+        protected override IEnumerable<OrderByField> GetDefaultOrderByFields()
+        {
+            yield return new OrderByField(new TableColumn(new Column("Id"), From.Source.Alias));
+        }
+
+        public override void AddWhere(UnaryWhereParameter whereParams)
+        {
+            WhereClauses.Add(new UnaryWhereClause(new AliasedWhereFieldReference(From.Source.Alias, new WhereFieldReference(whereParams.FieldName)),
```
droyad: You've built a nice immutable AST, but then suddenly here we fall back into a mutable list. I think we should go one step further and make a Query AST object that encapsulates the whole query (Select, From, Where and OrderBy). This single root object is then wrapped by the QueryBuilder to provide a nicer interface/backwards compatibility.

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
+        ISubquerySource AsSource();
+        ISubquerySourceBuilder<TRecord> Alias(string subqueryAlias);
+    }
+
+    public interface IJoinSourceQueryBuilder<TRecord> : IQueryBuilder<TRecord> where TRecord : class
+    {
+        IJoinSourceQueryBuilder<TRecord> On(string leftField, JoinOperand operand, string rightField);
+    }
+
+    public interface IQueryBuilder<TRecord> where TRecord : class
```
droyad: How hard would it be to add an immutable QueryBuilder?

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
+    {
         IQueryBuilder<TRecord> Where(string whereClause);
-        IQueryBuilder<TRecord> Where(string fieldName, SqlOperand operand, object value);
-        IQueryBuilder<TRecord> Where(string fieldName, SqlOperand operand, object startValue, object endValue);
-        IQueryBuilder<TRecord> Where(string fieldName, SqlOperand operand, IEnumerable<object> values);
-        
+        IQueryBuilder<TRecord> WhereParameterised(string fieldName, UnarySqlOperand operand, Parameter parameter);
```
droyad: Why change the name of this method? Isn't it going to make it harder to update all the code that uses it?

I'm also not sure what you mean by `Parameterised` as `Where(string)` can also (and usually is) parameterised

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
+        IQueryBuilder<TRecord> WhereParameterised(string fieldName, BinarySqlOperand operand, Parameter startValueParameter, Parameter endValueParameter);
+        IQueryBuilder<TRecord> WhereParameterised(string fieldName, ArraySqlOperand operand, IEnumerable<Parameter> parameterNames);
+
         IOrderedQueryBuilder<TRecord> OrderBy(string orderByClause);
         IOrderedQueryBuilder<TRecord> OrderByDescending(string orderByClause);
+        IQueryBuilder<TRecord> IgnoreDefaultOrderBy();
```
droyad: Is this needed? Previously we had a default orderby (`ID`) that was used if the user didn't specify an `OrderBy`. This method implies the new behaviour that it always adds `ID` to the order by unless you call this method.

I can't think of a situation where you want to not have an order by

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
 
-        IQueryBuilder<TRecord> Join(IJoin join);
-        IQueryBuilder<TRecord> Parameter(string name, object value);
-        IQueryBuilder<TRecord> LikeParameter(string name, object value);
-        IQueryBuilder<TRecord> LikePipedParameter(string name, object value);
-        IQueryBuilder<TRecord> View(string viewName);
-        IQueryBuilder<TRecord> Table(string tableName);
-        IQueryBuilder<TRecord> Hint(string tableHint);
-        IQueryBuilder<TRecord> NoLock();
+        IQueryBuilder<TRecord> Column(string name);
```
droyad: Since this is interface most likely encountered by the consumer, some XML doc should be added. I'm unsure at first glance what this method does.

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
+        IQueryBuilder<TRecord> Column(string name, string columnAlias);
+        IQueryBuilder<TRecord> Column(string name, string columnAlias, string tableAlias);
+        IQueryBuilder<TRecord> AllColumns();
```
droyad: Isn't this the default? What does this method end up being used for?

source/Nevermore/IQueryBuilder.cs
```diff
@@ -1,40 +1,85 @@
...
         bool Any();
         TRecord First();
         IEnumerable<TRecord> Take(int take);
         List<TRecord> ToList(int skip, int take);
         List<TRecord> ToList(int skip, int take, out int totalResults);
         List<TRecord> ToList();
-        void Delete();
         IEnumerable<TRecord> Stream();
         IDictionary<string, TRecord> ToDictionary(Func<TRecord, string> keySelector);
+        Parameters Parameters { get; }
```
droyad: This and the property below are mutable. Is the intention that the consumer add their own parameters directly to these collections?

source/Nevermore/IQueryBuilder.cs
droyad: Adding the lambda syntax to this method would be super useful `query.Column(q => q.Name)`

source/Nevermore/QueryBuilderWhereExtensions.cs
```diff
@@ -1,40 +1,58 @@
...
+using System.Collections;
+using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
+using Nevermore.AST;
 
 namespace Nevermore
 {
+    public enum SqlOperand
```
droyad: Should this be removed as it has been superseded? It should be easy to find-replace instances of this.

source/Nevermore/UnaryWhereParameter.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Collections.Generic;
```
droyad: Filename doesn't reflect that there are multiple classes here. Split out each class into a file (preferred) or rename the file.

source/Nevermore/TableAliasGenerator.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public class TableAliasGenerator : ITableAliasGenerator
+    {
+        int tableCount = 0;
+
+        public string GenerateTableAlias(string tableName = null)
+        {
+            // Return a predictable alias so we don't blow out SQL's query-plan cache.
+            tableCount++;
+            return $"ALIAS_{AliasLabel(tableName)}_{tableCount}";
```
droyad: I can't find any tests that show the SQL generated by this

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.ShouldGenerateSelectForComplicatedSubqueryJoin.approved.txt
```diff
@@ -1,4 +1,3 @@
-SELECT t1.* FROM (SELECT * FROM dbo.[Orders]) t1
-INNER JOIN (SELECT * FROM dbo.[Customers] WHERE (IsActive = 1)) t2 ON t1.[CustomerId] = t2.[Id] AND t1.[Owner] = t2.[Owner]
-INNER JOIN (SELECT * FROM dbo.[Accounts] WITH (UPDLOCK)) t3 ON t1.[AccountId] = t3.[Id]
-ORDER BY t1.[Id]
\ No newline at end of file
+SELECT t2.* FROM dbo.[Orders] t2
+INNER JOIN (SELECT * FROM dbo.[Customers] WHERE (IsActive = 1) ORDER BY [Id]) t1 ON t2.[CustomerId] = t1.[Id] AND t2.[Owner] = t1.[Owner]
```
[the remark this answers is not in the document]
TomPeters: In the test method, this subquery is generated by:

```
            var customers = CreateQueryBuilder<IDocument>("Customers")
                .Where("IsActive = 1")
                .OrderBy("Id");
```

So the order by is only added to this subquery because the consumer explicitly added it there. I can rename it to something other than `Id` though to make it clearer that it is not coming from the default order by clause.

source/Nevermore/IQueryBuilder.cs
TomPeters: I think that would be very easy :)

source/Nevermore/IQueryBuilder.cs
TomPeters: One of the aims is to be able to to use this query builder for stored procs / functions. Which means `Parameter` is a little bit overloaded. It means
1. The `@` parameter in the sql query string
2. The actual value associated with the `@` parameter that gets used at execution time

For 1. we need a way of parameterizing the sql query without providing a value (existing code makes you always provide a value). The new method `WhereParameterised` allows you to specify a where clause with a parameter but __without__ a value.

Any existing usages will switch to using extension methods, so don't need to change at all, and these extension method are implemented by:
1. Calling `WhereParameterised` (which allows us to share the same code for parameterising where clause in stored procs and normal queries)
2. Adding the actual parameter value to the `CommandParameterValues` collection

I imagine that it is mostly Stored Procs and Functions that would use `WhereParameterised` directly instead of the `Where` extension methods, so one reason that I named this method `WhereParameterised` was to make it distinct from `Where` and discourage people from using it in normal queries.

source/Nevermore/IQueryBuilder.cs
TomPeters: I believe that some of our stored procs do not have a default order by. For example, the dashboard query.

source/Nevermore/IQueryBuilder.cs
TomPeters: Yes this is the default, but only if you haven't specified anything else. So if you want to select all columns (i.e. `*`) and also something else (eg `Row_Number()`), then you need to explicitly specify that you want `AllColumns`.

Kind of the same as order by: if you want to order by `Foo` then by `Id`, you still need to explicitly add the `.OrderBy("Id")` even though `Id` is the default order by.

source/Nevermore/IQueryBuilder.cs
TomPeters: No, consumers should not add directly to these collections. I exposed these because I need them to be public in order to create stored procs / views / functions in extension methods. I'm thinking that maybe these should actually be instance methods on the query builder to avoid `Parameters` being made public. Alternatively, I can expose `Parameters` immutably.

source/Nevermore/QueryBuilderWhereExtensions.cs
TomPeters: Yes I only left it for backwards compatibility. Happy to remove it though.

source/Nevermore/TableAliasGenerator.cs
TomPeters: We are mocking out the `ITableAliasGenerator` in `QueryBuilderFixture`. I presume this was done before because it used to generate random strings in some cases and we needed it to be predictable for the tests.

I'll change this to use the concerete `TableAliasGenerator` so that it covers this logic as well.

source/Nevermore/SelectBuilder.cs
TomPeters: That's basically what the `ISelectBuilder` is. It's not part of the AST. It has mutable objects that represent Where, OrderBy and Select. From is a slightly more complicated so happens to be immutable. 

Maybe I'm not understanding you correctly.

source/Nevermore/IQueryBuilder.cs
droyad: You could make it `internal` if the extension methods are in the same assembly.

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.ShouldGenerateSelectForComplicatedSubqueryJoin.approved.txt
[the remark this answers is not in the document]
droyad: Ah right, nevermind then

source/Nevermore/IQueryBuilder.cs
TomPeters: True, I suppose that's the easiest option

TomPeters: Actually this doesn't work. It is just an interface, so you can't make it internal.

source/Nevermore.Tests/Nevermore.Tests.csproj
```diff
@@ -1,5 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
```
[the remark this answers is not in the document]
TomPeters: Done

source/Nevermore/Nevermore.csproj
```diff
@@ -1,5 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
```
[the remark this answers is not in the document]
TomPeters: Done

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.ShouldGenerateSelectForComplicatedSubqueryJoin.approved.txt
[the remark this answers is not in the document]
TomPeters: I have changed the field name

source/Nevermore/AST/Select.cs
TomPeters: Done. I only added `ToString` - this felt the most efficient summary of the properties on each AST node anyway, so I didn't think anything more was required in most cases.

source/Nevermore/IQueryBuilder.cs
TomPeters: Done

source/Nevermore/TableAliasGenerator.cs
TomPeters: Done

source/Nevermore/UnaryWhereParameter.cs
TomPeters: Done

source/Nevermore/IQueryBuilder.cs
droyad: Should we add it from the get go?

source/Nevermore/SelectBuilder.cs
droyad: Perhaps I missed it in the code, but is there an object in the AST that represents the Query itself? i.e I can get a reference to that object and see the whole query end to end?

source/Nevermore/IQueryBuilder.cs
TomPeters: Personally I would rather add it later to limit the scope of work. It is a non breaking change so can be in a minor version release. But open to being convinced otherwise.

source/Nevermore/IQueryBuilder.cs
droyad: `Where(string fieldName, SqlOperand operand, object value)` and `Where(string fieldName, SqlOperand operand, IEnumerable<object> values)` should quite commonly used since we introduced them. We've been encouraging people to steer away from `Where(string)` as the former can be strongly typed and automatically hook up the right parameters, eg `Where(nameof(Project.Name), SqlOperand.Equals, "foo")`

source/Nevermore/SelectBuilder.cs
TomPeters: Yep, the root of the AST is the `Select` class.

source/Nevermore/IQueryBuilder.cs
droyad: They are rather the exception as the Sprocs/views do ordering internally

droyad: Also if there is no ordering currently, I see no harm in ordering by `ID`

source/Nevermore/QueryBuilderWhereExtensions.cs
droyad: Probably for the better to reduce confusion given the ease and reliability of conversion

source/Nevermore/IQueryBuilder.cs
TomPeters: Fair enough, I'll remove this :+1:

source/Nevermore/IQueryBuilder.cs
TomPeters: Yep, and I haven't made any changes to that. People can still create Where expressions with `Where(string)` if they want to shoot themselves in the foot, or they can use extension methods like `Where(string fieldName, SqlOperand operand, object value)`

source/Nevermore/QueryBuilderWhereExtensions.cs
TomPeters: :+1: The only problem is `Contains`, `StartsWith` and `EndsWith` which was some sugar that we added on top of sql (change the way we formatted the parameter values, rather than changing the actual sql query). I should be able to migrate this to using the new linq expression extension methods though.

source/Nevermore/IQueryBuilder.cs
TomPeters: I'm thinking of leaving these as is. Moving stored procs / views / functions to instance methods doesn't mean that I can remove these, because there are other places that need them. For example, you still need to copy them to a new query builder by accessing them through the public `IQueryBuilder` interface when you add a join to your query.

The other option (changing the types exposed here to be immutable) doesn't feel like it adds a lot of value to me because we are already cloning the objects when you access these public properties. So as a consumer, you are unable to affect the query by mutating these objects.

source/Nevermore/QueryBuilderWhereExtensions.cs
TomPeters: Done

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.ShouldGenerateSelectForComplicatedSubqueryJoin.approved.txt
```diff
@@ -1,4 +1,3 @@
...
+INNER JOIN (SELECT * FROM dbo.[Accounts] WITH (UPDLOCK)) t3 ON t2.[AccountId] = t3.[Id] ORDER BY t2.[Id]
```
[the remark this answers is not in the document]
TomPeters: Done. I added quite a bit of formatting (not just to order by) so everything should read much nicer now.

source/Nevermore/IQueryBuilder.cs
TomPeters: Done

source/Nevermore/IQueryBuilder.cs
TomPeters: Done

TomPeters:
Added a new breaking change to the description:
Order By column names are always escaped in sql. Previously you could do OrderBy("[Id]") but now this is invalid, because it will be double escaped in sql (this same code will now end up as ORDER BY [[Id]])

TomPeters:
I've updated Octopus server here: https://gist.github.com/TomPeters/2864429f7da7c2dddaf870b34a59fe1b

TomPeters:
I've updated octopus server here: https://github.com/OctopusDeploy/OctopusDeploy/pull/1940 and from my manual testing, everything appears to be working correctly and all (non-flaky) tests are passing.

I did find a couple of issues during testing that I have resolved. The good news is that these issues were really obvious and generally broke lots of things, so were easy to discover, and were also caught by tests.

TomPeters:
I had broken the following scenario:

```
var foo = transaction.Query<Foo>().Where(...);
var bar = transaction.Query<Bar>().Where(...)
    .Join(foo.Subquery()).On(...);
```

You would expect the generated query to be something like
```
SELECT * FROM (SELECT * FROM dbo.[Foo] WHERE ...) ALIAS_GENERATED_1
INNER JOIN (SELECT * FROM dbo.[Bar] WHERE...) ALIAS_GENERATED_2 ON ...
```

Except instead, both of the subqueries were being generated with the same alias: `ALIAS_GENERATED_1`. This of course caused an execution time error.

The reason was because each of these subqueries originated from their own `QueryBuilder`, each of which had its own instance of `TableAliasGenerator`. And each instance of `TableAliasGenerator` has state that it relies on to generate unique aliases.

So to fix this, I have moved the instance of `TableAliasGenerator` to instead be stored on the `RelationalTransaction`. Which means that if you were to now do:

```
transaction.Query<Foo>().Subquery().ToList();
transaction.Query<Bar>().Subquery().ToList();
```

Then this would generate the following two sql queries:
`SELECT * FROM (SELECT * FROM dbo.[Foo]) ALIAS_GENERATED_1`
`SELECT * FROM (SELECT * FROM dbo.[Bar]) ALIAS_GENERATED_2`

So the queries generated within the same transaction could affect each other in terms of the aliases in each query. This feels ok to me, because we should generally be generating the same set of queries for the same type of transaction, so the aliases should still be consistent across different transactions doing the same type of operation.

TomPeters:
@droyad Could you have another look at this PR now? I think I've address all of your original comments, fixed any bugs that I've found and integrated it into Octopus Server and done some testing. Everything appears to be working nicely as far as I can tell.

TomPeters:
[Nevermore refactor](https://trello.com/c/Ast1kpxM/71-nevermore-refactor)

PR #40, opened 2018-03-06
Added support for loading more than 2100 ids in the Load() method

droyad:


droyad:
Yeh, I copied it from there. Didn't want to clash with the one in Shared.

PR #41, opened 2018-03-08
allow subclasses from column data, as well as JSON

MichaelJCompton:
Means that subtype mappings can pull from different columns as well as JSON, so allowing document hierarchies that pull from different bits of the table as well as different bits of JSON.

### DIFF
--- a/source/Nevermore/Mapping/SubTypedResolver.cs
+++ b/source/Nevermore/Mapping/SubTypedResolver.cs
@@ -64,10 +64,7 @@ namespace Nevermore.Mapping
         public override Func<IDataReader, Type> TypeResolverFromReader(Func<string, int> columnOrdinal)
         {
             var colIndex = columnOrdinal(column.ColumnName);
-            return reader => GetType(
-                0 <= colIndex && colIndex < reader.FieldCount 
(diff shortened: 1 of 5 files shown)


source/Nevermore/Mapping/SubTypedResolver.cs
```diff
@@ -64,7 +64,10 @@ public override Type GetTypeFromInstance(object instance)
         public override Func<IDataReader, Type> TypeResolverFromReader(Func<string, int> columnOrdinal)
         {
             var colIndex = columnOrdinal(column.ColumnName);
-            return reader => GetType(reader[colIndex].ToString());
+            return reader => GetType(
+                0 <= colIndex && colIndex < reader.FieldCount 
+                ? reader[colIndex].ToString() 
+                : "");
```
MichaelJCompton: E.g. in Octopus we check if an upgrade is compatible with license before kicking it off, but that requires loading things from DB.  But new Machines code can't query cause the new code won't work with Machine types cause the type columns aren't in the DB till the upgrade happens.

This essentially says, use the current type, rather than resolving a subtype, if the data to resolve on isn't found.

PR #42, opened 2018-03-22
Allow the same parameter to be specified multiple times in the same query

TomPeters:
I will probably not end up merging this to master, but instead re-do the changes on top of the changes in https://github.com/OctopusDeploy/Nevermore/pull/39

For now, this branch gives us a pre-release version to work with in the `spaces` branch in octopus server, until we can make the change on top of the AST branch once it has merged to master.


Flamage82:
We'll need to adjust this approach to handle multiple where clauses on the same column, eg:
`WHERE Name != 'Foo' AND Name != 'Bar'`

PR #45, opened 2018-04-17
Added a bit of extra type safety around unique parameters

TomPeters:


source/Nevermore/ParameterNameGenerator.cs
```diff
@@ -1,16 +1,17 @@
...
 {
-    public interface IParameterNameGenerator
+    public interface IUniqueParameterGenerator
     {
-        string GenerateUniqueParameterName(string parameterDescription);
+        UniqueParameter GenerateUniqueParameterName(Parameter parameter);
     }
 
-    class ParameterNameGenerator : IParameterNameGenerator
+    class UniqueParameterGenerator : IUniqueParameterGenerator
```
Flamage82: If you wanted to get really protective, you can restrict the UniqueParameter class so it is only able to be created inside the UniqueParameterGenerator.  https://stackoverflow.com/questions/1664793/how-to-restrict-access-to-nested-class-member-to-enclosing-class.  See my PR here: https://github.com/OctopusDeploy/Nevermore/pull/46

Issue #52, opened 2018-06-15
Add support for automatically maintaining a Related Documents table


droyad:
At the moment the only way to determine whether a document (child) is used by another document (parent) is to query all known parent tables. 

Since almost all relationships (in Octopus) are many-to-many, this lookup needs to be done on a `RelatedDocumentsId` column, which contains pipe delimited child document ids, or do a `Like` clause on the `JSON` column. On a large database, the two lookup methods take about 500ms and 2,000ms respectively.

This feature adds the automatic maintenance of one or more `RelatedDocument` tables containing the many-to-many relationships. e.g. if the document `Doc-1` references `Rel-1` and `Rel-2`, after an insert or update, the table would contain:
```
Id       RelatedDocumentId
-----------------------------
Doc-1    Rel-1
Doc-1    Rel-2
```

A lookup on this table based on `RelatedDocumentId` in the example above (500,000 rows) takes about 3ms. If expanded to all relationships, I estimate this table about be about 10-100 million rows, which should still be quite quick.

This feature does support splitting the relations across many tables.

PR #53, opened 2018-06-15
Related Document table support

droyad:
image

See #52 for details on this feature

The first commit is a refactor of how we generate Insert/Update/Delete scripts with the addition of tests to make sure the next commit does not regress this.

The second commit is the meat of the feature, adding the mapping and extra statements required.


source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.InsertMultipleDocumentWithManyRelatedDocuments.approved.txt
```diff
@@ -0,0 +1,23 @@
+INSERT INTO dbo.[TestDocumentWithRelatedDocuments]  (AColumn, Id, JSON) VALUES 
+(@0__AColumn, @0__Id, @0__JSON)
+,(@1__AColumn, @1__Id, @1__JSON)
+,(@2__AColumn, @2__Id, @2__JSON)
+INSERT INTO [RelatedDocuments] ([Id], [RelatedDocumentId]) VALUES
+(@0__Id, @relateddocuments_0)
```
michaelnoonan: I like how the parameters are used efficiently.

droyad:
@michaelnoonan Yeh, lets talk it out next week. Something to mull over till then, in some cases, particularly variables, we are not certain as to the type of the variable, or even if it is a valid variable. One solution is to perhaps ignore those variable that are not typed, or linked to well known variable named (eg `Octopus.Account.Id`)

michaelnoonan:
I'm wondering if we should delegate the responsibility of defining the
reference table to the mapping, and delegate collection of the references
to the model itself?

One upside of this approach: developers will be more inclined to pay
attention when they add new references to a document, rather than having to
realise they need to add another chunk of code into the document map. I
don't normally need to update the map when I add a reference to a document
since those references would normally just live inside the JSON. Like why
is it feels like the document itself is the most likely place already have
special type information like the variable scenario you mentioned before:
where variables have a much better idea about the types they can be linked
to.

Either way we could probably add a convention test that would remind people
to consider document references whenever they add something new to a
document.

pawelpabich:
A lot of changes here :) but if I understand correctly there is nothing that would prevent us from referencing missing data here? If so then that would have to be done in Octopus when we create new tables using foreign keys?

Would it make sense to make this solution aware of Spaces? If it was, we could created FKs that also include `SpaceId` and possibly prevent users from referencing valid documents from other spaces.

droyad:
@pawelpabich It might be clearer if you view the change without the first commit (https://github.com/OctopusDeploy/Nevermore/pull/53/files/ce804edbb2ce7fc545d569ccf07afe7feec19364..55d8e7d109fb23c448a9aa55928a47fc40e200dc)

There is nothing to enforce referential integrity, that is still up to the application. The main driver for this change is speed of querying.

droyad:
@TomPeters It would be interesting to see how the new permission checks would use this table. At the moment I want to concentrate on the problem at hand unless it paints us into a corner that is prohibitive to get out of (ie it is difficult to Spaces or column metadata). Also remember that Spaces may not be a DAL concern.

pawelpabich:
@droyad that's what I thought and it make sense you are only focusing on performance at Nevermore level.  

That being said this is a pretty powerful tool and my worry is that without some additional safety net (maybe at the Octopus level) we can actually increase the number of problems with have with corrupted data.

droyad:
@pawelpabich Happy to hear ideas on how we could achieve this. The main stumbling block are VariableSets.

PR #54, opened 2018-06-29
Store the table name instead of the document type

droyad:
A previous change changed the `IdTypeColumnName` to store the table name instead of the type and I didn't rename the column correctly at that time, which was misleading. After implementing in Octopus, it turns out determining the real `RelatedDocumentType` is tricky and not really required, just storing the table name is sufficient.

### DIFF
--- a/source/Nevermore/IRelationalStore.cs
+++ b/source/Nevermore/IRelationalStore.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Data;
 using System.Text;
+using Nevermore.Mapping;
 
 namespace Nevermore
 {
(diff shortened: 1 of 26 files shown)


droyad:
[Store the table name instead of the document type](https://trello.com/c/LCcIpt1j/109-store-the-table-name-instead-of-the-document-type)

source/Nevermore/IRelationalStore.cs
```diff
@@ -11,5 +12,7 @@ public interface IRelationalStore
         IRelationalTransaction BeginTransaction(RetriableOperation retriableOperation = RetriableOperation.Delete | RetriableOperation.Select, string name = null);
         IRelationalTransaction BeginTransaction(IsolationLevel isolationLevel, RetriableOperation retriableOperation = RetriableOperation.Delete | RetriableOperation.Select, string name = null);
         void WriteCurrentTransactions(StringBuilder sb);
+        DocumentMap GetMappingFor<T>();
+        DocumentMap GetMappingFor(Type type);
```
johnsimons: This one could be an extension method to keep the interface slim ?

droyad: Not without making the mapping object public. Also it'd be a PITA to mock

johnsimons: I actually meant the `DocumentMap GetMappingFor<T>();`

droyad: They forward to different methods (as confusing as that is :/)

johnsimons: Should this be a TryGetMappingFor ?

PR #56, opened 2018-07-05
Change subtype handling

slewis74:
We can now support using a JSON converter to handle subtyping the same way we do in the API. This means the type namespaces won't be serialized into the DB and we are not going to break if we move types to different namespaces/assemblies.

Added tests to ensure the serialization is correct, and to guard against it accidentally changing in the future.

### DIFF
--- a/source/Nevermore/Serialization/InheritedClassConverter.cs
+++ b/source/Nevermore/Serialization/InheritedClassConverter.cs
@@ -120,7 +120,7 @@ namespace Nevermore.Serialization
         {
             if (!DerivedTypeMappings.ContainsKey(derivedType))
             {
-                throw new Exception($"Unable to determine type to deserialize. {TypeDesignatingPropertyName} does not map to a known type");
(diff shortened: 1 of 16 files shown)


source/Nevermore/Serialization/InheritedClassConverter.cs
```diff
@@ -0,0 +1,130 @@
...
+                .Where(p => p.CanWrite))
+            {
+                var val = jo.GetValue(prop.Name);
+                if (val != null)
+                    prop.SetValue(instance, val.ToObject(prop.PropertyType, serializer), null);
+            }
+            return instance;
+        }
+
+        protected virtual TypeInfo GetTypeInfoFromDerivedType(string derivedType)
```
droyad: This confused me a bit, perhaps it would be clearer if you had a seperate class for the enum case and left this abstract. I think then you can also do `where TDerived : Enum`

slewis74: 👍

PR #57, opened 2018-07-16
Adding support for specifying timeout for IDeleteQueryBuilder

johnsimons:


source/Nevermore.Tests/Delete/DeleteQueryBuilderFixture.cs
```diff
@@ -27,7 +27,7 @@ public void ShouldGenerateDelete()
         {
             string actual = null;
 
-            void ExecuteDelete(Type _, Where where, CommandParameterValues __) => actual = where.GenerateSql().Trim();
+            void ExecuteDelete(Type _, Where where, CommandParameterValues __, int? ___) => actual = where.GenerateSql().Trim();
```
michaelnoonan: Can't they all be `_`? Instead of `_` then `__` then `___`? I thought the compiler magicked these away?

PR #59, opened 2018-07-27
Memoized the property reflection to shave a few more milliseconds off the serialization

droyad:


droyad:
Less worried about the API. This is mostly for the table cache item cloning.

PR #63, opened 2018-11-12
Optimisation for Any linq.

johnsimons:
Instead of using Count > 0 we now use EXISTS.

Why we use EXISTS, read https://stackoverflow.com/a/11130604/90882


source/Nevermore/QueryBuilder.cs
```diff
@@ -230,8 +230,10 @@ public int Count()
         public bool Any()
         {
             var clonedSelectBuilder = selectBuilder.Clone();
-            clonedSelectBuilder.AddColumnSelection(new SelectExistsSource());
-            var result = transaction.ExecuteScalar<int>(new ExistsSelectBuilder(clonedSelectBuilder).GenerateSelect().GenerateSql(), paramValues);
+            clonedSelectBuilder.RemoveOrderBys();
+            clonedSelectBuilder.IgnoreDefaultOrderBy();
+            var query = new IfExpression(new ExistsExpression(clonedSelectBuilder.GenerateSelect()), new SelectConstant("1"), new SelectConstant("0")).GenerateSql();
```
distantcam: Eww, magic strings

source/Nevermore/AST/CustomExpression.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Nevermore.AST
+{
+    public class CustomExpression : IExpression
```
distantcam: This class feels like a "do whatever you want" class. Do we need this? If we're going to force an expression structure for the SQL generation then we shouldn't have this class. Otherwise why not just use this for all SQL generation?

TomPeters: It's a back door (that we currently have btw!) to allow you to do things that nevermore doesn't yet support. It is just a more specific backdoor than we currently have :) Ideally we wouldn't have this backdoor at all but that involves more work to support a bunch of extra features. I don't think we have the capacity to do this work right now... better to move partially in the right direction then not at all.

>Otherwise why not just use this for all SQL generation?

There's static typing benefits whereever we can use our non-custom expression classes.

source/Nevermore/QueryBuilder.cs
TomPeters: This code seems appropriate to me. What would you prefer to see instead?

distantcam: A `const` for true and false, which is then also used in the return below.

source/Nevermore/AST/SelectConstant.cs
```diff
@@ -0,0 +1,17 @@
...
+        readonly string constant;
+
+        public SelectConstant(string constant)
+        {
+            this.constant = constant;
+        }
+
+        public string GenerateSql()
+        {
+            return $"SELECT {constant}";
```
distantcam: What if the `constant` is a string? Also could this be a vector for an injection attack?

source/Nevermore/QueryBuilder.cs
TomPeters: Sure, thought you might have been thinking along different lines, that works :+1:

johnsimons:
@TomPeters @distantcam there was a small problem we forgot the `@`, see https://github.com/OctopusDeploy/Nevermore/pull/63/commits/d2f29790d71e42d84b199ddd8372053cc5d12b45

But now i think it is all done ?

TomPeters:
@johnsimons minor, we have just changed the implementation of `.Any()` to behave the same way but more efficiently.

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderStateFixture.cs
```diff
@@ -50,8 +50,11 @@ public void ShouldNotModifyQueryBuilderStateFromAny()
 
             queryBuilder.Any();
 
-            LastExecutedQuery().ShouldBeEquivalentTo(@"SELECT COUNT(*)
-FROM dbo.[Accounts]");
+            LastExecutedQuery().ShouldBeEquivalentTo(@"IF EXISTS(SELECT *
+FROM dbo.[Accounts])
+    SELECT @true_0
```
droyad: Why are you doing this instead of `SELECT Cast(1 as bit)`?

PR #64, opened 2018-12-05
Reduce unnecessary nesting of unions

TomPeters:
Prevent unnecessary nested of subqueries when unioning multiple queries.

Previously, if you did a bunch of unions, like

```
builder.Query<Foo>()
.Union(builder.Query<Bar>()) // line 2
.Union(builder.Query<Baz>())
```

This would generate sql like:

```
select * from (
    select * from Foo
    union all
    select * from Bar) ALIAS_GENERATED_1
union all
select * from Baz
```

This is because after line 2 in the above query, you could add extra statements like Where, order by, column selection, etc. At this point, it would make sense to apply these statements to the things that you have so far unioned. The only way to do this is to wrap up everything that you have so far unioned into a subquery so that you can apply these modifications to the subquery. For example if you did:

```
builder.Query<Foo>()
.Union(builder.Query<Bar>())
.Where(a => a.Name == "Bob")
.Union(builder.Query<Baz>())
```

This would generate sql like:
```
select * from (
    select * from Foo
    union all
    select * from Bar) ALIAS_GENERATED_1
where Name = 'Bob'
union all
select * from Baz
```

The code was naive in that it would always wrap whatever you had in a subquery just in case you decided to add these modifications.

This PR is a bit smarter and only wraps the sql in a subquery if you have made any modifications. If you have not made any modifications, then it does not wrap it in a subquery. So for the first example, the generated sql should now look like:

```
select * from Foo
union all
select * from Bar
union all
select * from Baz
```


source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.ShouldGenerateMultipleUnionsWithoutNesting.approved.txt
```diff
@@ -0,0 +1,12 @@
+SELECT *
```
NickJosevski: Yea good compared to how it would have been

```

SELECT *
FROM (
    SELECT *
    FROM (
        SELECT [Id] AS [Id]
        FROM dbo.[Orders]
        UNION
        SELECT [Id] AS [Id]
        FROM dbo.[Account]
    ) ALIAS_GENERATED_1
    UNION
    SELECT [Id] AS [Id]
    FROM dbo.[Customers]
) ALIAS_GENERATED_2
ORDER BY ALIAS_GENERATED_2.[Id]
```

PR #66, opened 2019-02-22
Allow instance instantiation with private constructor

tothegills:
Relates to OctopusDeploy/Issues#5325

### DIFF
--- a/source/Nevermore/RelationalTransaction.cs
+++ b/source/Nevermore/RelationalTransaction.cs
@@ -461,7 +461,7 @@ namespace Nevermore
                         }
                         else
                         {
-                            instance = (T) Activator.CreateInstance(instanceType);
+                            instance = (T) Activator.CreateInstance(instanceType, true);
                         }
 
                         var specificMapping = mappings.Get(instance.GetType());


tothegills:
[Tests demonstrating reporting endpoints work #3496](https://trello.com/c/iFyk8N6U/192-tests-demonstrating-reporting-endpoints-work-3496)

source/Nevermore/RelationalTransaction.cs
```diff
@@ -461,7 +461,7 @@ IEnumerable<T> Stream<T>(IDbCommand command, DocumentMap mapping)
                         }
                         else
                         {
-                            instance = (T) Activator.CreateInstance(instanceType);
+                            instance = (T) Activator.CreateInstance(instanceType, true);
```
johnsimons: @tothegills could this be considered a breaking change?

tothegills: Yes I agree, there could be private constructors that are now preferred for instance creation.

johnsimons: So, how do we move forward, should this be a flag in Nevermore that by default if off?

PR #69, opened 2019-03-27
Provide RawSqlQuery method on RelationalTransaction

Waldo000000:
- Nevermore's SQL graph is great, but it's also not 100% complete
  - An example in this case: `SELECT MaxFoo = MAX(Foo) OVER (PARTITION BY Bar)`. Currently not supported.
- For tricky little queries, sometimes we just want to spin up an `IQueryBuilder` from raw SQL, while still being able to take advantage of all the Nevermore goodness (e.g. document mappings, the composition of other operations like filters, etc)
- This draft PR will introduce a method `transaction.RawSqlQuery(...)` to allow for exactly this
- I imagine in future, these raw SQL usages _could_ later be migrated to use first class Nevermore SQL objects if/when they become available. This PR just gives us a bit more flexibility to stay unblocked by some edge cases in the meantime.
- [x] TODO: Tests

### DIFF
--- a/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
+++ b/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
@@ -265,6 +265,7 @@ namespace Nevermore.IntegrationTests.RelationalTransaction
                 trn.Commit();
 
                 var productSubtotalQuery =    @"SELECT
+                                                    Id,
                                                     ProductId,
                                                     ProductName,
                                                     Subtotal
--- a/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
+++ b/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
@@ -1,12 +1,24 @@
 ﻿using System.Linq;
 using FluentAssertions;
 using Nevermore.IntegrationTests.Model;
+using Nevermore.Mapping;
 using NUnit.Framework;
 
 namespace Nevermore.IntegrationTests.RelationalTransaction
 {
     public class LoadFixture : FixtureWithRelationalStore
     {
+        public override void SetUp()
+        {
+            base.SetUp();
+
+            Mappings.Install(
+                new DocumentMap[]
+                {
+                    new ProductSubtotalMap()
+                }
+            );
+        }
 
         [Test]
         public void LoadWithSingleId()
@@ -22,7 +34,7 @@ namespace Nevermore.IntegrationTests.RelationalTransaction
         {
             using (var trn = Store.BeginTransaction())
             {
-                trn.Load<Product>(new[] {"A", "B"});
+                trn.Load<Product>(new[] { "A", "B" });
             }
         }
 
@@ -43,7 +55,7 @@ namespace Nevermore.IntegrationTests.RelationalTransaction
                 trn.LoadStream<Product>(Enumerable.Range(1, 3000).Select(n => "ID-" + n));
             }
         }
-        
+
         [Test]
         public void StoreNonInheritedTypesSerializesCorrectly()
         {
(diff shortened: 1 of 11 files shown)


Waldo000000:
@droyad I've added an integration test (with a somewhat gnarly little query, and some parameters as suggested) and I've marked this PR ready for review. Keen to get your feedback 🙏  :-)

source/Nevermore/IQueryExecutor.cs
```diff
@@ -76,6 +76,14 @@ public interface IQueryExecutor
         /// <returns>A stream of resulting documents.</returns>
         ITableSourceQueryBuilder<TDocument> TableQuery<TDocument>() where TDocument : class, IId;
 
+        /// <summary>
+        /// Returns strongly typed documents from the specified raw SQL query.
+        /// </summary>
+        /// <typeparam name="TDocument">The type of document being queried. Results from the database will be mapped to this type.</typeparam>
+        /// <param name="query">The SQL query to execute. Example: <c>SELECT COUNT(*) FROM...</c></param>
+        /// <returns>A builder to further customize the query.</returns>
+        ISubquerySourceBuilder<TDocument> RawSqlQuery<TDocument>(string query) where TDocument : class, IId;
```
Waldo000000: I'd love to remove the constraint that `TDocument : IId`. In the integration test, we were forced to set a dummy `Id` on `ProductSubtotal`, but it's not used. I'm having a look at whether we can remove this constraint; input welcome.

droyad: Yeh it doesn't make sense in this context

source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
```diff
@@ -240,5 +252,70 @@ public void StoreAndLoadStringInheritedTypesThatAreProperties()
...
+                                                        [Name] ProductName,
+                                                        CAST(JSON_VALUE([JSON], '$.Price') AS decimal) Price
+                                                        FROM Product
+                                                    ) p ON p.ProductId = l.ProductId
+                                                    GROUP BY p.ProductId, p.ProductName
+                                                ) p3
+                                                WHERE Subtotal >= @MinimumSubtotal";
+
+                var doubleDigitUnicornProductSubtotals = trn.RawSqlQuery<ProductSubtotal>(productSubtotalQuery)
+                    .Where(s => s.ProductName.Contains("Unicorn"))
```
droyad: That's an interesting one. How does it know to add an `AND` instead of `WHERE`

source/Nevermore/IQueryExecutor.cs
Waldo000000: Thanks @droyad. The `IId` constraint was there because we add [an `ORDER BY [Id]` clause](https://github.com/OctopusDeploy/Nevermore/blob/master/source/Nevermore/SelectBuilder.cs#L177) by default in some cases.

I took a stab at removing the constraint along with the default `ORDER BY` clause in [this commit](https://github.com/OctopusDeploy/Nevermore/pull/69/commits/4eb4a2626725a5302016df7c2ec4468a2c85c85d), but I'll revert it for now because there are some locations in Octopus server where [we seem to rely](https://github.com/OctopusDeploy/OctopusDeploy/blob/master/source/Octopus.Server/Web/Api/Actions/EnvironmentsMachinesResponder.cs#L86) on this default ordering behavior. Too risky for too little gain for me right now.

droyad: Ah yeh, we definitely need the default ordering on most queries (for paging and other things as you have found).

source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
Waldo000000: The raw SQL is wrapped up as a subquery, so the outer `Where` applies to that:
```
SELECT *
FROM (
    ...
    WHERE Subtotal >= @MinimumSubtotal
) ALIAS_GENERATED_1
WHERE ([ProductName] LIKE @productname_0)
```

PR #70, opened 2019-03-28
 Ability to compare to null in where expressions

markledwich2:
Added ability to compare to null in where expressions
e.g. `t.TableQuery<Customer>().Where(c => c.Nickname == null)`

### DIFF
--- a/source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
+++ b/source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
@@ -16,7 +16,7 @@ namespace Nevermore.IntegrationTests
     public class IntegrationTestDatabase
     {
         readonly TextWriter output = Console.Out;
-        readonly string SqlInstance = "(local)\\SQLEXPRESS,1433";
+        readonly string SqlInstance = Environment.GetEnvironmentVariable("NevermoreTestServer") ?? "(local)\\SQLEXPRESS,1433";
         readonly string TestDatabaseName;
         readonly string TestDatabaseConnectionString;
 
--- a/source/Nevermore/QueryBuilder.cs
+++ b/source/Nevermore/QueryBuilder.cs
@@ -44,9 +44,15 @@ namespace Nevermore
             return this;
         }
 
-        public IQueryBuilder<TRecord> WhereNull(string fieldName, bool not)
+        public IQueryBuilder<TRecord> WhereNull(string fieldName)
         {
-            selectBuilder.AddWhere(new IsNullWhereParameter(fieldName, not));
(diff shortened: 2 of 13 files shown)


source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
```diff
@@ -16,7 +16,7 @@ namespace Nevermore.IntegrationTests
     public class IntegrationTestDatabase
     {
         readonly TextWriter output = Console.Out;
-        readonly string SqlInstance = "(local)\\SQLEXPRESS,1433";
+        readonly string SqlInstance = Environment.GetEnvironmentVariable("NevermoreTestServer") ?? "(local)\\SQLEXPRESS,1433";
```
droyad: I'm curious about this addition. Was it you have a default instance and it didn't get picked up by .NET Core? If so we have the `SqlAlias` library for that.

source/Nevermore/QueryBuilder.cs
```diff
@@ -44,6 +44,12 @@ public IQueryBuilder<TRecord> Where(string whereClause)
             return this;
         }
 
+        public IQueryBuilder<TRecord> WhereNull(string fieldName, bool not)
```
droyad: I would add `WhereNotNull` instead of passing a `not` flag

markledwich2: Yes, that's better

source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
markledwich2: The /SQLEXPRESS alias redirection is esoteric. I have been ensuring that database connection dependencies are configurable through normal app settings or environment variables everywhere.

source/Nevermore/QueryBuilderWhereExtensions.cs
```diff
@@ -146,7 +146,7 @@ static IQueryBuilder<TRecord> AddUnaryWhereClauseFromExpression<TRecord>(IQueryB
             var fieldName = property.Name;
             
             if (value == null && new[] {UnarySqlOperand.Equal, UnarySqlOperand.NotEqual}.Contains(operand))
-                return queryBuilder.WhereNull(fieldName, operand == UnarySqlOperand.NotEqual);
+                return operand == UnarySqlOperand.NotEqual ? queryBuilder.WhereNotNull(fieldName) : queryBuilder.WhereNull(fieldName);
```
droyad: Can I be pedantic and say inverting this to the positive would read better?

"If equal operator then call where null else call where not null" vs "If not equal operator then call where not null else call where null"

PR #71, opened 2019-04-11
Added ReadOnly option for columns

markledwich2:
For some columns, the database itself calculates values, and Nevermore should be able to be configured to read a column but not include in insert/update statements. RowVersion and SQL Calculated Columns are examples of this. 

For example, in a derived class of  `DocumentMap` , this configures the RowVersion column to be read from the database and not updated/inserted:
```
Column(m => m.RowVersion).ReadOnly();
```

### DIFF
--- a/source/Nevermore.IntegrationTests/RelationalStoreFixture.cs
+++ b/source/Nevermore.IntegrationTests/RelationalStoreFixture.cs
@@ -198,5 +198,18 @@ namespace Nevermore.IntegrationTests
                 ex.Message.Should().Be("Customers must have a unique name");
             }
         }
+
+        [Test]
+        public void ShouldHanldeRowVersionColumn()
+        {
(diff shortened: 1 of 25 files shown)


markledwich2:
@droyad this is the same as when we went through together, but with an extra test for the RowVersion scenario.

source/Nevermore.IntegrationTests/RelationalStoreFixture.cs
```diff
@@ -198,5 +198,18 @@ public void ShouldShowFriendlyUniqueConstraintErrors()
                 ex.Message.Should().Be("Customers must have a unique name");
             }
         }
+
+        [Test]
+        public void ShouldHanldeRowVersionColumn()
```
droyad: Spelling

Issue #68, opened 2019-03-14
Add way to provide command timeout for IQueryBuilder Fetch and Stream methods


markledwich2:
For some queries (e.g. fetching all rows from a large table), the default 60s timeout is not long enough. 

Add a way to provide command timeout for `IQueryBuilder` Fetch and Stream methods. 

The workaround at the moment is to use the following instead:
`transaction.ExecuteReader<TDocument>(query.GetSelectBuilder().GenerateSelect().GenerateSql(), query.ParameterValues, commandTimeout);`

PR #73, opened 2019-07-22
Add more options to specify command timeouts and add missing timing log messages

flin-8:
Add missing timing log messages for `RelationalTransaction.ExecuteReader()` and `RelationalTransaction.Stream()`

Add missing command timeout options to `IQueryExecutor` methods

Add `IQueryBuilder.WithTimeout()` method which allows you to specify a command timeout when using `IQueryBuilder`.

Fixes: https://github.com/OctopusDeploy/Nevermore/issues/68

### DIFF
--- a/source/Nevermore/IQueryBuilder.cs
+++ b/source/Nevermore/IQueryBuilder.cs
@@ -76,6 +76,13 @@ namespace Nevermore
     public interface IQueryBuilder<TRecord> where TRecord : class
     {
         /// <summary>
+        /// Sets the command timeout for the query execution
+        /// </summary>
+        /// <param name="commandTimeout">A custom timeout to use for the command instead of the default.</param>
+        /// <returns>The query builder that can be used to further modify the query, or execute the query</returns>
+        IQueryBuilder<TRecord> WithTimeout(TimeSpan commandTimeout);
+
+        /// <summary>
         /// Adds a custom where expression to the query. Avoid using this as it is difficult to refactor. Prefer using Where methods from <see cref="QueryBuilderWhereExtensions" />
         /// </summary>
         /// <param name="whereClause">Where clause expression that will be inserted directly into the resulting SQL string</param>
--- a/source/Nevermore/IQueryBuilder.cs
+++ b/source/Nevermore/IQueryBuilder.cs
@@ -73,14 +73,14 @@ namespace Nevermore
         IJoinSourceQueryBuilder<TRecord> On(string leftField, JoinOperand operand, string rightField);
     }
 
-    public interface IQueryBuilder<TRecord> where TRecord : class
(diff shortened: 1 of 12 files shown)


droyad:
Also, this is a breaking API change, so a major version bump is needed (`+semver:major` commit comment)

flin-8:
> There are a heap of unrelated changes that look like it's applying a line length limit. I don't think that limit is part of our standard code style. This is making it hard to review.
> 
> I think `commandTimeout` should be `TimeSpan?`

OK Will revert those.

Would you like me to change the existing command timeouts to `TimeSpan`'s too?

droyad:
Ah I didn't see existing ones. It should at least be consistent. With the `int` it's unclear whether it is seconds or milliseconds. (I guess seconds knowing SQL)

droyad:
@TomPeters might have thoughts

TomPeters:
I think we should change all timeouts to `TimeSpan`.

Also, when you are using the query builder, perhaps it would be better for the usage to be like:

```
transaction.Query<Document>().WithTimeout(TimeSpan.FromSeconds(10)).ToList()`
```

droyad:
@zentron looks like there will be a breaking change, good timing to bump the client libs as well. I think we should drop netstandard 1.5 while we are at it.

source/Nevermore/IQueryBuilder.cs
```diff
@@ -75,6 +75,13 @@ public interface IJoinSourceQueryBuilder<TRecord> : IQueryBuilder<TRecord> where
 
     public interface IQueryBuilder<TRecord> where TRecord : class
     {
+        /// <summary>
+        /// Sets the command timeout for execution of the query
+        /// </summary>
+        /// <param name="commandTimeout">A custom timeout to use for the command instead of the default.</param>
+        /// <returns>The query builder that can be used to further modify the query, or execute the query</returns>
+        IQueryBuilder<TRecord> WithTimeout(TimeSpan commandTimeout);
```
TomPeters: This implementation allows you to do things like
```
transaction.Query<Customer>().InnerJoin(
    transaction.Query<People().WithTimeout(TimeSpan.FromSeconds(5))
).On(...).ToList()
```

In this case the Timeout will be ignored, which is kind of confusing.

Perhaps a better option might be to only allow you to invoke methods that execute the query after you have called `WithTimeout(...)`. So something like

```
public interface ICompleteQuery<TRecord>
{
    IEnumerable<TRecord> Stream();
    List<TRecord> ToList();
    //...all the other methods that actually executed the query go here
}
```
```
public interface IQueryBuilder<TRecord> : ICompleteQuery<TRecord>
{
    ICompleteQuery<TRecord> WithTimeout(TimeSpan commandTimeout);
}
```

flin-8: Gotcha!

PR #78, opened 2019-08-12
Order by with table alias

NickJosevski:
To minimize the number of subqueries required to generate what's required by `Replace_LatestSuccessfulDeployments` the orderby column requires alias support so it can be selected from another table in the query

### DIFF
--- a/source/Nevermore/QueryBuilder.cs
+++ b/source/Nevermore/QueryBuilder.cs
@@ -203,7 +203,7 @@ namespace Nevermore
         
         public IOrderedQueryBuilder<TRecord> OrderBy(string fieldName, string tableAlias)
         {
-            selectBuilder.AddOrder(fieldName, tableAlias, true);
+            selectBuilder.AddOrder(fieldName, tableAlias, false);
             return this;
         }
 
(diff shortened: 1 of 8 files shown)


source/Nevermore/QueryBuilder.cs
```diff
@@ -200,13 +200,24 @@ public IOrderedQueryBuilder<TRecord> OrderBy(string fieldName)
             selectBuilder.AddOrder(fieldName, false);
             return this;
         }
-
+        
+        public IOrderedQueryBuilder<TRecord> OrderBy(string fieldName, string tableAlias)
+        {
+            selectBuilder.AddOrder(fieldName, tableAlias, true);
```
TomPeters:
```suggestion
            selectBuilder.AddOrder(fieldName, tableAlias, false);
```

source/Nevermore/SelectBuilder.cs
```diff
@@ -260,6 +260,11 @@ public virtual void AddOrder(string fieldName, bool @descending)
         {
             OrderByClauses.Add(new OrderByField(new Column(fieldName), @descending ? OrderByDirection.Descending : OrderByDirection.Ascending));
         }
+        
+        public virtual void AddOrder(string fieldName, string tableAlias, bool @descending)
```
TomPeters:
```suggestion
        public void AddOrder(string fieldName, string tableAlias, bool @descending)
```

PR #79, opened 2019-08-12
Rename `First` (which can return null) to `FirstOrDefault`

Waldo000000:
## Background
`ICompleteQuery.First` currently returns `null` if the result of the query returns no records. In contrast, the commonly used `Enumerable.First` method never returns null, and the behavior of `ICompleteQuery.First` more closely mirrors that of `Enumerable.FirstOrDefault`. This inconsistency has [caused some confusion](https://github.com/OctopusDeploy/Octofront/pull/1093#discussion_r311825078), which gave rise to the suggestion for the following change in Nevermore.

## Fix
This PR marks the `ICompleteQuery.First` as `Obsolete`, and introduces a new, equivalent method in its place called `ICompleteQuery.FirstOrDefault`, which is more consistent with the naming scheme used by `Enumerable` and is hoped to be less surprising for users.

## Impact
While this PR will likely initially cause a significant number of `Obsolete` warnings to be emitted for most projects due to existing usages of `First`, these warnings should at least be trivial to fix. To review this PR, please assess whether you agree that the potential benefit in clarity/maintainability for users justifies the impact of this change.

### DIFF
--- a/source/Nevermore/ICompleteQuery.cs
+++ b/source/Nevermore/ICompleteQuery.cs
@@ -18,7 +18,7 @@ namespace Nevermore
         /// <returns>Returns true if there are any rows in the result set, otherwise false</returns>
         bool Any();
 
-        [Obsolete("Use FirstOrDefault instead.")]
+        [Obsolete("Use FirstOrDefault instead")]
         TRecord First();
         
         /// <summary>
--- a/source/Nevermore/ICompleteQuery.cs
+++ b/source/Nevermore/ICompleteQuery.cs
@@ -18,11 +18,13 @@ namespace Nevermore
         /// <returns>Returns true if there are any rows in the result set, otherwise false</returns>
         bool Any();
 
-        /// <summary>
-        /// Executes the query and retuns the first row
-        /// </summary>
-        /// <returns>The first row in the result set</returns>
+        [Obsolete("Use FirstOrDefault instead.")]
         TRecord First();
+        
+        /// <summary>
+        /// Executes the query and returns the first row, or null if there are no rows 
+        /// </summary>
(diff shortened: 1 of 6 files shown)


source/Nevermore/ICompleteQuery.cs
```diff
@@ -18,11 +18,13 @@ public interface ICompleteQuery<TRecord> where TRecord : class
         /// <returns>Returns true if there are any rows in the result set, otherwise false</returns>
         bool Any();
 
+        [Obsolete("Use FirstOrDefault instead")]
```
pawelpabich: `Use FirstOrDefault instead` might not be enough for someone to understand why then need to make the change..  

Should we extend the explanation to something like this:
`First returns the first row, or null if there are no rows To make your code easier to read use FirstOrDefault  instead.`

TomPeters: Instead of making this `Obsolete`, should we just remove this instead? Saves having two methods around for a while, and saves having to come back and remove this later (if we decide to). I don't _think_ there are many people other than us that use Nevermore. When we have made breaking changes in the past, I don't think anyone has complained.

Waldo000000: Good feedback, thanks :+1: I've thought a bit about this and I think the benefit from the explanation/hint for upgrading users (rather than just a broken build) outweighs the cost of having two methods around for a while (which seems pretty harmless, given nice IDE treatment of `Obsolete` methods) and the cost of having to come back and remove this later (which is optional, and trivial).

I agree we might be fine to make this is a breaking change, but it's just _so easy_ to make it non-breaking in this case :-)

PR #81, opened 2019-09-04
Improve error message when a document is missing a setter

Waldo000000:
## Steps to reproduce
```
public class Customer : IId
{
    public string Id { get; }
}

// ...

transaction.Insert(new Customer());
```

## Actual result
Confusing null ref:
```
System.NullReferenceException: Object reference not set to an instance of an object.
```

## Expected result
A slightly more informative error message that points to the issue (i.e. in this case, that the `Id` has no `set` accessor):
```
System.InvalidOperationException: Cannot write to a property without a setter
```

### DIFF
--- a/source/Nevermore/Nevermore.csproj
+++ b/source/Nevermore/Nevermore.csproj
@@ -29,7 +29,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Nevermore.Contracts\Nevermore.Contracts.csproj" />
-    <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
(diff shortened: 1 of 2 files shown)


source/Nevermore/Nevermore.csproj
```diff
@@ -29,6 +29,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Nevermore.Contracts\Nevermore.Contracts.csproj" />
+    <PackageReference Include="JetBrains.Annotations" Version="2019.1.3" />
```
Waldo000000: _I'm happy to remove this if you think the cost of taking on this dependency outweighs the benefits it may provide._

droyad: Yeh I think so. If anything we'd use the new C# 8.0 feature

PR #84, opened 2019-10-06
Improve development experience of Nevermore on Linux

TomPeters:
Enables development on non-windows machines.

Changes include
- Removing .net framework target support. Only target netstandard and netcore
- Add support for supplying a username/password instead of relying on integrated auth in integration tests.

### DIFF
--- a/source/Nevermore.Contracts/Nevermore.Contracts.csproj
+++ b/source/Nevermore.Contracts/Nevermore.Contracts.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <Description>A JSON Document Store library for SQL Server</Description>
     <Authors>Octopus Deploy</Authors>
-    <TargetFrameworks>netstandard1.2;net451</TargetFrameworks>
     <AssemblyName>Nevermore.Contracts</AssemblyName>
     <PackageId>Nevermore.Contracts</PackageId>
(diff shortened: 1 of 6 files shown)


johnsimons:
@TomPeters So we dropping `net451`? 
This is an open-source lib, is everyone ok with it?

TomPeters:
This lib will still work with `net461` and above, so we are only dropping `net451 and net46`.

Internal discussion: https://octopusdeploy.slack.com/archives/CDANN5QLT/p1574819442198900

I think we want to keep moving this project forward, so dropping `net451` at some point is desirable.

source/Nevermore.Contracts/Nevermore.Contracts.csproj
```diff
@@ -21,6 +20,7 @@
     <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">True</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
     <WarningsAsErrors />
+    <TargetFramework>netstandard1.2</TargetFramework>
```
johnsimons: Should we also include `netstandard2.0`?

TomPeters: Not explicitly - by targeting `netstandard1.2` we should be forwards compatible with all of the same frameworks that are compatible with `netstandard2.0`

PR #87, opened 2019-12-12
Typed strings: support tiny string types on documents

Waldo000000:
## Context
The primitive type, `string`, is very useful for modeling all sorts of information, for example, IDs, names, free-form text, etc. However, using `string` itself to model all different “types” of string-like information in our applications can cause us some problems, as will be discussed below. This PR addresses these problems by enabling us to (optionally) use different types of string to model different kinds of information.

## Problem

For example, a function with this signature:

```
public void Foo (string barId, string bazId)
```

...can be called as follows:

```
Foo(someBazId, someBarId)
```

At first glance, this seems fine. But did you spot the incorrect `string, string` parameter order in the call there? Even the compiler does not alert us to any problem here.

Wouldn’t it be nice if the compiler could help us out with this and catch this problem early, without us wasting time with a long feedback loop, or worse, a bug going out to production? Wouldn’t it be nice to get some help from our IDEs, without needing to squint and carefully double check parameter order?

## Previous workarounds
There are a few possible approaches to getting the C# type system to address this problem and give us these benefits.

Occasionally, we have seen people reach for the following convention of passing entire (strongly typed) objects, rather than just the property that will be used:

```
public void Foo (Bar bar, Baz baz)
```

This has the disadvantages of sometimes unnecessarily loading entire objects (perhaps even with an extra roundtrip to the DB), as well as passing a thicker interface than required to the dependency. For this reason, such changes are sometimes short-lived, and can cause churn in the codebase.

## Solution

This PR attempts to sidestep the above drawbacks, and achieve a sort of “best of both worlds”, i.e. only passing around what each abstraction needs (e.g. often, just an ID), but also achieving better type safety by specifying a type that maps 1-to-1 with our domain concept (not just a vague, catch-all `string` type), as in:

```
public void Foo (BarId barId, BazId bazId)
```

## This PR
This PR adds support for non-string-literal property types in Nevermore documents, including `Id`.

A common use case for passing around `strings` in our applications is when we're passing IDs of documents. Nevermore has been modified in this PR to allow the `Id` property of documents to be more strongly typed to a specific string type.

These changes should be backward-compatible through just adding overloads and additional generic `IId` and `IDocument` interfaces (small caveat: theoretically, adding overloads is [not always](https://codeblog.jonskeet.uk/2018/04/13/backward-compatibility-and-overloading/) backward-compatible, but I think we’re ok in this case).

In the future, I imagine we’ll deprecate and remove the previous style. We probably could have made this in one big breaking change, but I personally prefer shipping smaller incremental changes.

## How to review this PR

- Upgrade your application to this version of Nevermore, and modify an existing document to additionally implement `IDocument<TId>`, and/or change some property on one of your documents to use a new specific implementation of `IIdWrapper`, rather than just `string`. Verify all Nevermore functions should still work as expected.

## PR dependencies

- [x] https://github.com/OctopusDeploy/TinyTypes/pull/1
- This PR will unblock the following: 
  - [ ] https://github.com/OctopusDeploy/OctopusDeploy/pull/5392
  - [ ] https://github.com/OctopusDeploy/Octofront/pull/1444


source/Nevermore.IntegrationTests/TypedStringConverter.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            if (reader.TokenType != JsonToken.String)
+                throw new JsonSerializationException(
+                    $"Unexpected token or value when parsing TypedString. Token: {reader.TokenType}, Value: {reader.Value}");
+            return Activator.CreateInstance(objectType, (string) reader.Value);
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(TypedString);
```
flin-8: Will type equality work when you have a concrete `TypedString` ? `IsAssignableFrom()`?

Waldo000000:
[Typed Strings](https://trello.com/c/6Y1MHzDb/1-typed-strings)

source/Nevermore.IntegrationTests/TypedStringConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            if (reader.TokenType != JsonToken.String)
+                throw new JsonSerializationException(
+                    $"Unexpected token or value when parsing TypedString. Token: {reader.TokenType}, Value: {reader.Value}");
+            return Activator.CreateInstance(objectType, (string) reader.Value);
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            return typeof(TimeSpan).GetTypeInfo().IsAssignableFrom(objectType);
```
flin-8: `TimeSpan` eh? TypedTime anyone?

source/Nevermore.IntegrationTests/Legacy/QueryBuilderIntegrationFixture.cs
```diff
@@ -0,0 +1,61 @@
...
+                {
+                    new Customer {FirstName = "Alice", LastName = "Apple", Nickname = null},
+                    new Customer {FirstName = "Bob", LastName = "Banana", Nickname = ""},
+                    new Customer {FirstName = "Charlie", LastName = "Cherry", Nickname = "Chazza"}
+                })
+                    t.Insert(c);
+                t.Commit();
+                
+                var customersNull = t.TableQuery<Customer>()
+                    .Where(c => c.Nickname == null)
```
pawelpabich: It might be useful to convert `Nickname` to a `TypedString` to make sure LINQ to Nevermore works with the new library.

NuGet.Config
```diff
@@ -5,7 +5,8 @@
   </activePackageSource>
 
   <packageSources>
+    <add key="Feedz.io" value="https://packages.octopushq.com/dependencies/nuget/index.json" />
```
Waldo000000: This was introduced because the test projects take a dependency on the (currently private) `TinyTypes` package. Need a decision on whether we want that to remain the case, or make `TinyTypes` public.

source/Nevermore.IntegrationTests/Legacy/FixtureWithRelationalStore.cs
```diff
@@ -0,0 +1,57 @@
+using System;
+using Nevermore.Contracts;
+using Nevermore.Mapping;
+using NUnit.Framework;
+
+namespace Nevermore.IntegrationTests.Legacy
```
Waldo000000: Copied the old integration tests to a `Legacy` subdir here. We could remove them if/when we want to only cover the new style of documents.

source/Nevermore.IntegrationTests/Model/Brand.cs
```diff
@@ -1,16 +1,25 @@
...
-using Newtonsoft.Json;
+using TinyTypes.TypedStrings;
 
 namespace Nevermore.IntegrationTests.Model
 {
-    public abstract class Brand : IDocument
+    public abstract class Brand : IDocument<BrandId>, IId
     {
-        public string Id { get; protected set; }
+        public BrandId Id { get; protected set; }
```
Waldo000000: ✨

source/Nevermore.IntegrationTests/Model/Brand.cs
```diff
@@ -1,16 +1,25 @@
...
         public string Name { get; set; }
 
         public abstract string Type { get; }
 
         public string Description { get; set; }
+
+        string IId.Id => Id?.Value;
```
Waldo000000: This is ugly, but pragmatic. 

Here we explicitly implement the old non-generic interfaces (`IId`/`IDocument`) as well, which require a strictly `string` ID. This allows for reuse of the old succinct Nevermore signatures without adding type parameters where they don't provide value. This choice also happens to provide for backwards compatibility.

source/Nevermore.Tests/Util/LiteralStringIdDataModificationQueryBuilderFixture.cs
```diff
@@ -0,0 +1,444 @@
...
+using Nevermore.AST;
+using Nevermore.Contracts;
+using Nevermore.Mapping;
+using Nevermore.Util;
+using Newtonsoft.Json;
+using NUnit.Framework;
+
+namespace Nevermore.Tests.Util
+{
+    public class LiteralStringIdDataModificationQueryBuilderFixture
```
Waldo000000: This is just a rename of the previous `DataModificationQueryBuilderFixture`, using the old style (`string` ID) documents.

source/Nevermore.Contracts/IId.cs
```diff
@@ -4,4 +4,18 @@ public interface IId
...
+    public interface IId<out TId> where TId : IIdWrapper
+    {
+        TId Id { get; }
+    }
+
+    /// <summary>
+    /// Implementations must have a constructor that takes a single string value
+    /// <see cref="Nevermore.Mapping.AmazingConverter"/>
+    /// </summary>
+    public interface IIdWrapper
```
Waldo000000: Should this really be `IStringWrapper` (or even just `ITypedString`), so that it can be used just as easily with other properties as with IDs? We might be missing some test cases for documents with typed strings on non-ID properties.

source/Nevermore.Tests/Nevermore.Tests.csproj
```diff
@@ -32,5 +32,6 @@
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
     <PackageReference Include="FluentAssertions" Version="4.19.2" />
     <PackageReference Include="Castle.Core" Version="4.2.1" />
+    <PackageReference Include="TinyTypes" Version="0.1.0-roy-typed-string0008" />
```
Waldo000000: - [ ] Bump to released version

source/Nevermore.IntegrationTests/Nevermore.IntegrationTests.csproj
```diff
@@ -31,5 +31,6 @@
     <PackageReference Include="dbup" Version="4.1.0" />
     <PackageReference Include="Castle.Core" Version="4.2.1" />
     <PackageReference Include="TestStack.BDDfy" Version="4.3.2" />
+    <PackageReference Include="TinyTypes" Version="0.1.0-roy-typed-string0008" />
```
Waldo000000: - [ ] Bump to released version

source/Nevermore.Contracts/IId.cs
flin-8: Does Nevermore need to know about typed strings in other properties? that's probably not Nevermore's concern?

flin-8: Actually we missed a crucial case: allowing custom types in columns (if a property maps to a column), it blows up currently: https://github.com/OctopusDeploy/Nevermore/blob/f9fbe8a87bcc79382b5154f50727b8fdeb402169/source/Nevermore/Mapping/DatabaseTypeConverter.cs#L9

flin-8: Ah you can fix it by calling
`Column(m => m.NotProductId, c => c.DbType = DbType.String);`

source/Nevermore/Util/TinyTypeExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+                .GetInterfaces()
+                .SingleOrDefault(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(ITinyType<>));
+        }
+
+        public static object GetTinyTypeValue(this Type tinyType, object value)
+        {
+            if (value == null) 
+                throw new ArgumentNullException(nameof(value));
+            
+            var property = tinyType.GetProperty(nameof(ITinyType<object>.Value), BindingFlags.Instance | BindingFlags.Public);
```
flin-8: Slightly annoying that we need to reflect on this type every time (unless we have also have a non-generic `ITinyType` interface)

Waldo000000: Ah, interesting. This annoyance is required to support `int` (or other value types), right? If we were only supporting reference types, we'd get away with `value is ITinyType<object>`:

```
void Main()
{
	ContributeParameter(new TypedString("foo")); // WORKS: Unboxed value without reflection: foo
	ContributeParameter(new TypedInt(42));       // :-(    Unboxed value without reflection: UserQuery+TypedInt
	ContributeParameter("just a string");        // WORKS: Unboxed value without reflection: just a string
	ContributeParameter(100);                    // WORKS: Unboxed value without reflection: 100
}

void ContributeParameter(object value)
{
	var unboxedValue = value is ITinyType<object> tinyType ? tinyType.Value : value;

	$"Unboxed value without reflection: {unboxedValue}".Dump();
}

interface ITinyType<out T>
{
	T Value { get; }
}

class TypedString : ITinyType<string>
{
	public TypedString(string value) { Value = value; }
	public string Value { get; private set; }
}

class TypedInt : ITinyType<int>
{
	public TypedInt(int value) { Value = value; }
	public int Value { get; private set; }
}
```

flin-8: >  If we were only supporting reference types, we'd get away with value is ITinyType<object>

Correct...your thoughts on this?

Waldo000000: It's mainly been strings (particularly IDs) that have caught my eye to this point, but supporting value types would be neat. I think the reflection is ok 👍

source/Nevermore.IntegrationTests/Model/LineItem.cs
```diff
@@ -1,14 +1,30 @@
...
+        public LineItemId Id { get; private set; }
         public string Name { get; set; }
         public DateTime PurchaseDate { get; set; }
         public string ProductId { get; set; }
-        public int Quantity { get; set; }
+        public Quantity Quantity { get; set; }
+        string IId.Id => Id?.Value;
+    }
+
+    public class Quantity : TypedInt, Contracts.ITinyType<int>
```
Waldo000000: @flin-8 This boilerplate will be quite common, i.e. defined once per individual tiny type. If we were to aim to remove the double inheritance, i.e.:
```
public class Quantity : TypedInt
```

...how might we go about that, what implications would that have, and would it be worth it, to simplify this boilerplate?

(my assumption: Nevermore itself - not just its integration tests - would take a dependency on `Octopus.TinyTypes`, and we'd only have one `ITinyType<T>` interface)

flin-8: Mind you, you only need one boilerplate per type:
` public class NMTypedInt : TypedInt, Contracts.ITinyType<int>`
`public class Quantity : NMTypedInt`

This is still not ideal.

The question is are we OK with forcing Nevermore users to take a dependency on `Octopus.TinyTypes`

Waldo000000: >are we OK with forcing Nevermore users to take a dependency on Octopus.TinyTypes

I believe we discussed this at kick-off and we felt that the costs of breaking this dependency (i.e. requiring every consumer to carefully inherit from both interfaces, whether directly or indirectly) would outweigh the cost (e.g. of potential dependencies on two different versions of Newtonsoft).

Waldo000000: Correct me if my memory is wrong here @flin-8 :-)

source/Nevermore.IntegrationTests/Model/Brand.cs
Waldo000000: Good idea via @TomPeters to avoid this explicit `IId.Id` implementation (please correct me if my summary is wrong though Tom): only implement `IId`, and remove the `string Id` property and `IIdWrapper` entirely - rely on the IdColumn mapping to read/write to the `Id` property appropriately. 

I'll give this a go and see if I hit any sharp corners.

source/Nevermore.IntegrationTests/Legacy/QueryBuilderIntegrationFixture.cs
Waldo000000: I've done this 👍 It returns the expected results in the usual case...but with one limitation. Right now, these two filters both work equivalently:
- `.Where(c => c.Nickname == new Nickname("Chazza"))` 
- `.Where(c => c.Nickname == new SomeOtherTypedString("Chazza"))`

Both return documents that have a `Nickname` with value `"Chazza"`. It would be nicer if the latter always returned no results.

This behavior appears to be because command parameter values are unwrapped and eventually happily passed to the `SqlClient.SqlParameterCollection` as plain `strings`.

https://github.com/OctopusDeploy/Nevermore/blob/bd7ab2445ef4f2ac82882e8b04bff5563eb03b0f/source/Nevermore/CommandParameterValues.cs#L90

Waldo000000: Fixed in https://github.com/OctopusDeploy/Nevermore/pull/87/commits/4589bfeeebcc2a2a851342e180313e05f3cd9741. 

A bit heavy-handed maybe? It affects all queries, not just ones involving tiny types, and would now return no results for any query of the form:
```
Where(c => c.SomeProperty == someValueThatCannotBeAssignedToThatPropertyType)
```

NuGet.Config
Waldo000000: `TinyTypes` will already publish to NuGet for non-prerelease versions, just as `Nevermore` does (https://github.com/OctopusDeploy/TinyTypes/blob/master/build.cake#L111).

source/Nevermore.IntegrationTests/Model/Brand.cs
Waldo000000: I've pushed this change here: https://github.com/OctopusDeploy/Nevermore/pull/87/commits/35f57a50b798d711c88138cd69b9bd81827ebf96. Keen for feedback.

source/Nevermore.IntegrationTests/Legacy/QueryBuilderIntegrationFixture.cs
Waldo000000: Changed my mind, and reverted the above in https://github.com/OctopusDeploy/Nevermore/pull/87/commits/e5c79c6be0fb94ac34acc23aa0dd7741073268c0, so now the SQL filter generated from expressions like `.Where(c => c.Nickname == new SomeOtherTypeOfStringString("Chazza"))` will once again just do a basic SQL comparison on the underlying value.

Reasoning:
1. Both behaviors are potentially surprising (i.e. returning rows matching the underlying value, or returning no rows)
2. Neither behaviors achieve the goal of this work any more than the other, i.e. improving _compile-time_ type safety
3. We can't possibly hope to implement all custom `==` operator implementations in SQL
3. Given that, let's be honest with ourselves and just treat these expressions as syntactic sugar for generating SQL queries which, of course, end up just comparing serialized values

source/Nevermore.IntegrationTests/Model/Brand.cs
Waldo000000: Unsurprisingly, this breaks a _bunch_ of usages of `IId.Id` in `OctopusDeploy` (WIP in https://github.com/OctopusDeploy/OctopusDeploy/pull/5392).

PR #88, opened 2020-02-16
Added ability to change the default setting for MARS

pawelpabich:
Nevermore used to force MARS. We need a way to a) disable this behaviour and b) keep the current behaviour in place of existing consumers.

Use 
image to review the changes

### DIFF
--- a/source/Nevermore/RelationalStore.cs
+++ b/source/Nevermore/RelationalStore.cs
@@ -114,7 +114,7 @@ namespace Nevermore
         {
             var connectionStringBuilder = new SqlConnectionStringBuilder(connectionString)
             {
-                MultipleActiveResultSets = true,
+                //MultipleActiveResultSets = true,
                 ApplicationName = applicationName,
             };
 
--- a/source/Nevermore/RelationalStore.cs
+++ b/source/Nevermore/RelationalStore.cs
@@ -24,12 +24,10 @@ namespace Nevermore
         readonly ISqlCommandFactory sqlCommandFactory;
         readonly RelationalMappings mappings;
         readonly Lazy<RelationalTransactionRegistry> registry;
-        readonly Lazy<RelationalTransactionRegistry> registryWithMarsOn;
         readonly JsonSerializerSettings jsonSettings = new JsonSerializerSettings();
         readonly IRelatedDocumentStore relatedDocumentStore;
         readonly IKeyAllocator keyAllocator;
         readonly ObjectInitialisationOptions objectInitialisationOptions;
-        readonly bool enableMarsByDefault;
 
         /// <summary>
         ///
@@ -42,7 +40,7 @@ namespace Nevermore
         /// <param name="relatedDocumentStore"></param>
         /// <param name="keyBlockSize"></param>
         /// <param name="objectInitialisationOptions"></param>
-        /// <param name="enableMarsByDefault">MARS: https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/using-multiple-active-result-sets-mars?view=sql-server-ver15</param>
+        /// <param name="forceMars">MARS: https://docs.microsoft.com/en-us/sql/relational-databases/native-client/features/using-multiple-active-result-sets-mars?view=sql-server-ver15</param>
         public RelationalStore(
             string connectionString,
             string applicationName,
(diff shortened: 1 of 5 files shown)


source/Nevermore/RelationalStore.cs
```diff
@@ -37,7 +51,8 @@ public class RelationalStore : IRelationalStore
             JsonSerializerSettings jsonSettings,
             IRelatedDocumentStore relatedDocumentStore,
             int keyBlockSize = 20,
-            ObjectInitialisationOptions objectInitialisationOptions = ObjectInitialisationOptions.None)
+            ObjectInitialisationOptions objectInitialisationOptions = ObjectInitialisationOptions.None,
+            bool setMarsAsDefault = true)
```
andyinaus: May consider rename it to `isMarsEnabledByDefault or `enableMarsByDefault` as `setMarsAsDefault` sounds like a action.

pawelpabich: 👍

source/Nevermore/RelationalStore.cs
```diff
@@ -71,18 +87,25 @@ public class RelationalStore : IRelationalStore
...
             this.mappings = mappings;
             keyAllocator = new KeyAllocator(this, keyBlockSize);
 
             this.jsonSettings = jsonSettings;
             this.relatedDocumentStore = relatedDocumentStore;
             this.objectInitialisationOptions = objectInitialisationOptions;
+            this.setMarsAsDefault = setMarsAsDefault;
         }
 
         public string ConnectionString => registry.Value.ConnectionString;
```
andyinaus: I believe we will have to make the **Lazy registry** smarter as switching between different lazy factories is going to cause problems for code like the above `public string ConnectionString => registry.Value.ConnectionString;` as it depends on the **registry** not the other one. In other words, in the case of `registryWithMarsOn` in use, the above code and other code that depends on `registry` are no longer reflecting the actual registry instance that is in use any more. 

And I assume it's an expensive operation to initialize a `RelationalTransactionRegistry` instance (as we are using lazy initialization), initializing two instances may have some performance impact.

pawelpabich: > public string ConnectionString => registry.Value.ConnectionString;
👍 

> And I assume it's an expensive operation to initialize a RelationalTransactionRegistry instance (as we are using lazy initialization), initializing two instances may have some performance impact.

No, I don't think so.  It looks like we simply don't want to execute a delegate in the constructor.

source/Nevermore/RelationalStore.cs
```diff
@@ -71,24 +88,31 @@ public class RelationalStore : IRelationalStore
...
 
             this.jsonSettings = jsonSettings;
             this.relatedDocumentStore = relatedDocumentStore;
             this.objectInitialisationOptions = objectInitialisationOptions;
+            this.enableMarsByDefault = enableMarsByDefault;
         }
 
-        public string ConnectionString => registry.Value.ConnectionString;
-        public int MaxPoolSize => registry.Value.MaxPoolSize;
+        public string ConnectionString => SelectRegistry().ConnectionString;
```
andyinaus: This will only load the default one tho. It doesn't reflect the actual one in use and after having a closer look of this implementation, I believe initially this was designed to hold a **store level** registry and this registry is meant to have some sort of monitoring/logging helper ability of current used transactions in order to track the usage of the transaction/connection pool or gain the visibility. Thus, having two registries here will break the existing behavior if someone is using it as swapping between two registries will not reflect the actual number of transactions/connections in use in the store. e.g. `WriteCurrentTransactions(StringBuilder sb)` is broken now and each registry holds its own collection of transactions as well as its own sql connection pool due to different connection strings.

pawelpabich: There are two connections pools so having two registries actually makes a lot of sense. Each connection pool has its own limit.

That being said there is a bug here because `WriteCurrentTransactions` should use both registries.

andyinaus: Yeah, and the `MaxPoolSize` and `ConnectionString` property of `RelationalStore` are no longer valid as well. This `RelationalStore` was designed to be like **one-to-one** relationship between itself and `Connection Pool` as it has direct dependency on one single **connection string**.

PR #89, opened 2020-03-03
Log SQL parameters along with the query

flin-8:
Only logging the query doesn't give us the full picture, since we don't know what the parameter values are.

This will let us know the exactly query that was run on the machine to help diagnose performance issues.

### DIFF
--- a/source/Nevermore/RelationalTransaction.cs
+++ b/source/Nevermore/RelationalTransaction.cs
@@ -680,7 +680,7 @@ namespace Nevermore
 
         string QueryInfoForLogging(string query, CommandParameterValues parameterValues)
         {
-            return QueryInfoForLogging(query, parameterValues.Select(p => (p.Key, p.Value)).ToList());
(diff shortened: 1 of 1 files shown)


source/Nevermore/RelationalTransaction.cs
```diff
@@ -678,6 +678,25 @@ public override string ToString()
...
+        {
+            return QueryInfoForLogging(command.CommandText, command.Parameters.Cast<IDataParameter>().Select(p => (p.ParameterName, p.Value)).ToList());
+        }
+
+        string QueryInfoForLogging(string query, IReadOnlyList<(string ParameterName, object ParameterValue)> parameters)
+        {
+            if (!parameters.Any())
+                return query;
+
+            var parameterInfo = "Parameters:\r\n" + string.Join(",\r\n", parameters.Select(p => p.ParameterName + ": " + p.ParameterValue));
```
flin-8: Not sure if we should truncate this list so we only print the first X parameters (can have up to 2100)

flin-8:
Abandoning: https://octopusdeploy.slack.com/archives/C0K9DNQG5/p1583208531025500

PR #90, opened 2020-03-17
Ensure list parameter replacement doesn't partially match similar parameter names

adam-mccoy:
## Problem
This issue was discovered [here](https://github.com/OctopusDeploy/Issues/issues/6185).

When using a parameter with a list of values, Nevermore will rewrite the query to build an inline list featuring a new parameter for each list item. For example:

```
SELECT * FROM [Table] WHERE [Id] IN @ids
```
becomes:
```
SELECT * FROM [Table] WHERE [Id] IN (@ids_0, @ids_1, ...@ids_n)
```

This uses a Regex to match the name of the parameter to replace, but will also match other identifiers that partially match. For example:
```
SELECT * FROM [Table] WHERE [Id] IN @ids AND [OtherThing] = @ids123
```
becomes:
```
SELECT * FROM [Table] WHERE [Id] IN (@ids_0, @ids_1) AND [OtherThing] = (@ids_0, @ids_1)123
```
which is clearly not valid SQL.

## Solution
Adjust the Regex match pattern to include a look-ahead for a character that is not valid for use in identifiers `(letters, numbers, _, $ #, @)`, which would denote the end of the parameter name. This means that other identifiers which start with same characters would not be replaced.

### DIFF
--- a/source/Nevermore.Tests/CommandParameterValuesFixture.cs
+++ b/source/Nevermore.Tests/CommandParameterValuesFixture.cs
@@ -22,6 +22,22 @@ namespace Nevermore.Tests
             command.Parameters["someId_1"].Value.Should().Be("id2");
             command.Parameters["someIdentifier"].Value.Should().Be("value");
         }
+        
+        [Test]
+        public void ShouldReplaceParametersWithExactMatchEndOfQuery()
+        {
+            var parameters = new CommandParameterValues();
+            parameters.Add("someId", "value");
+            parameters.Add("someIdentifier", new[] { "id1", "id2" });
+            var command = new SqlCommand("SELECT * FROM [Table] WHERE [Id] = @someId AND [OtherId] IN @someIdentifier");
+
+            parameters.ContributeTo(command);
+
+            command.CommandText.Should().Be("SELECT * FROM [Table] WHERE [Id] = @someId AND [OtherId] IN (@someIdentifier_0, @someIdentifier_1)");
+            command.Parameters["someId"].Value.Should().Be("value");
+            command.Parameters["someIdentifier_0"].Value.Should().Be("id1");
+            command.Parameters["someIdentifier_1"].Value.Should().Be("id2");
+        }
 
         [TestCase("@")]
         [TestCase("#")]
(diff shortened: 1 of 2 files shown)


source/Nevermore.Tests/CommandParameterValuesFixture.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Nevermore.Tests
+{
+    public class CommandParameterValuesFixture
+    {
+        [Test]
+        public void ShouldReplaceParametersWithExactMatch()
+        {
+            var parameters = new CommandParameterValues();
+            parameters.Add("someId", new[] { "id1", "id2" });
+            parameters.Add("someIdentifier", "value");
```
flin-8: Should we also test the reverse? where `someId` is just one value and `someIdentifier` is an array?

source/Nevermore.Tests/CommandParameterValuesFixture.cs
```diff
@@ -0,0 +1,36 @@
...
+            var command = new SqlCommand("SELECT * FROM [Table] WHERE [Id] IN @someId AND [OtherId] = @someIdentifier");
+
+            parameters.ContributeTo(command);
+
+            command.CommandText.Should().Be("SELECT * FROM [Table] WHERE [Id] IN (@someId_0, @someId_1) AND [OtherId] = @someIdentifier");
```
flin-8: Assert the actual parameter values too

source/Nevermore.Tests/CommandParameterValuesFixture.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            parameters.ContributeTo(command);
+
+            command.CommandText.Should().Be("SELECT * FROM [Table] WHERE [Id] = @someId AND [OtherId] IN (@someIdentifier_0, @someIdentifier_1)");
+            command.Parameters["someId"].Value.Should().Be("value");
+            command.Parameters["someIdentifier_0"].Value.Should().Be("id1");
+            command.Parameters["someIdentifier_1"].Value.Should().Be("id2");
+        }
+
+        [TestCase("@")]
```
flin-8: I think it _might_ be cleaner if you just put the whole identifier  as the test case i.e. `@someId#entifier`

PR #92, opened 2020-04-06
Faster Load(array)

PaulStovell:
Nevermore provides a Load overload that takes an array of ID's. The common use case is that you perform a select, find a set of related document IDs, then load that set of related documents. 

It currently works by generating large "IN" statements. E.g.,:

    SELECT * FROM Product WHERE Id in (@ids_0, @ids_1, .....)

This has a few downsides:

- You can only have so many parameters in a statement, so when searching on hundreds of ID's, it needs to split them into chunks
 - Queries with different numbers of parameters (e.g., 317 vs 407) don't get to reuse the same query plan
 - It's generally just slower

This new version makes use of tabled value parameters. 

Here's the performance difference (on my local SQLEXPRESS)

Old approach:

```
Loaded 100 products in 42ms  // warm up
Loaded 100 products in 2ms 
Loaded 300 products in 32ms
Loaded 600 products in 79ms
Loaded 1200 products in 105ms
Loaded 1800 products in 130ms
Loaded 2400 products in 166ms
```

New approach (table valued parameters)

```
Loaded 100 products in 47ms
Loaded 100 products in 2ms
Loaded 300 products in 5ms
Loaded 600 products in 9ms
Loaded 1200 products in 16ms
Loaded 1800 products in 23ms
Loaded 2400 products in 30ms
```

The new approach is faster because SQL is able to reuse the same query plan whether you ask for 5 ID's or 5000. It also doesn't have to batch, so it can load thousands in one go. 

For large data sets, the difference is more obvious. Here it is loading 50,000 items:

```
Loaded 50000 products in 684ms (new way)
Loaded 50000 products in 7347ms (old way)
```

This feels fast enough (for 50,000 items!) that it might move this from being a strategy we want to avoid to one we can actually use more frequently. 

**There is one catch though**. You have to define this type (with this name):

```
CREATE TYPE dbo.[ParameterList] as TABLE([ParameterValue] nvarchar(300))
```

This would need to be a migration script in Octopus/Octofront.

### DIFF
--- a/source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
+++ b/source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
@@ -55,6 +55,7 @@ namespace Nevermore.IntegrationTests
         void CreateDatabase()
         {
             ExecuteScript(@"create database [" + TestDatabaseName + "] COLLATE SQL_Latin1_General_CP1_CS_AS", GetMaster(TestDatabaseConnectionString));
+            ExecuteScript(@"CREATE TYPE dbo.[ParameterList] as TABLE([ParameterValue] nvarchar(300))", TestDatabaseConnectionString);
         }
 
         void DropDatabase()
--- a/source/Nevermore/CommandParameterValues.cs
+++ b/source/Nevermore/CommandParameterValues.cs
@@ -2,10 +2,13 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Data;
+using System.Data.Common;
 #if NETFRAMEWORK
 using System.Data.SqlClient;
+using Microsoft.SqlServer.Server;
 #else
 using Microsoft.Data.SqlClient;
+using Microsoft.Data.SqlClient.Server;
 #endif
 using System.Linq;
 using System.Reflection;
@@ -50,6 +53,18 @@ namespace Nevermore
 
         public CommandType CommandType { get; set; }
 
+        public void AddTable(string name, IEnumerable<string> ids)
+        {
+            var idColumnMetadata = new SqlMetaData("ParameterValue", SqlDbType.NVarChar, 300);
+            
+            Add(name, ids.Where(v => !string.IsNullOrWhiteSpace(v)).Select(v =>
+            {
+                var record = new SqlDataRecord(idColumnMetadata);
+                record.SetValue(0, v);
+                return record;
+            }).ToList());
+        }
+        
         void AddFromParametersObject(object args)
         {
             if (args == null)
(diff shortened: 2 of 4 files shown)


PaulStovell:
A bit more about why it seems to be faster.

The giant "IN" clause generates a query plan like this:

image

While the table valued parameter approach looks like this:

image

The TVP approach only scans the parameter list - not the source table (which would likely be much larger).

tothegills:
[Table value parameters reference](https://docs.microsoft.com/en-us/sql/relational-databases/tables/use-table-valued-parameters-database-engine?view=sql-server-ver15)

source/Nevermore/CommandParameterValues.cs
```diff
@@ -139,4 +162,139 @@ public void AddRange(CommandParameterValues other)
             }
         }
     }
+
+    public class MyDataReader : DbDataReader
```
tothegills, line 166: What's the data reader for?

source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
```diff
@@ -55,6 +55,7 @@ public IntegrationTestDatabase()
         void CreateDatabase()
         {
             ExecuteScript(@"create database [" + TestDatabaseName + "] COLLATE SQL_Latin1_General_CP1_CS_AS", GetMaster(TestDatabaseConnectionString));
+            ExecuteScript(@"CREATE TYPE dbo.[ParameterList] as TABLE([ParameterValue] nvarchar(300))", TestDatabaseConnectionString);
```
tothegills, line 58: Nevermore includes a `KeyAllocator` upgrade script, should the creation of the `ParamerterList` type also be included as an "in-the-box" script?

source/Nevermore/RelationalTransaction.cs
```diff
@@ -116,22 +117,18 @@ public class RelationalTransaction : IRelationalTransaction
...
+            if (idList.Count == 0)
+                return new List<T>();
+            
+            var mapping = mappings.Get(typeof(T));
+            var tableName = mapping.TableName;
+            
+            var param = new CommandParameterValues();
+            param.AddTable("criteriaTable", idList);
+            return ExecuteReader<T>(
+                $"SELECT s.* FROM dbo.[{tableName}] s INNER JOIN @criteriaTable t on t.[ParameterValue] = s.[Id] order by s.[Id]",
```
tothegills, line 130: The `ORDER BY` caught my eye, I can't think of a horrible side effect but thought it worth calling out.

source/Nevermore/CommandParameterValues.cs
PaulStovell: Woops didn't mean to include that, thanks

source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
PaulStovell: Good idea, might add it as a second script

source/Nevermore/RelationalTransaction.cs
PaulStovell: Will drop it

PR #94, opened 2020-04-06
Support custom type definitions in the DB

slewis74:
Using a configuration object to coordinate the mappings and the custom types.

AmazingConverter is no longer static, it knows about custom types.

Key driver for this is to support Octopus using TinyTypes or extensible Enums concepts without having to make further changes to Nevermore itself.

This change makes #85 and #87 redundant, as they can be implemented by the consumer.


PaulStovell:
(I know this is a draft PR, but I had these thoughts as I read it - hopefully it's useful feedback)

Adding `RelationalStoreConfiguration` everywhere feels like a smell. It's not clear to me as a consumer of the API why I should need to pass this boilerplate everywhere. 

This part particularly feels very awkward:

```csharp
    var orderMap = new OrderMap(RelationalStoreConfiguration);
    RelationalStoreConfiguration.AddDocumentMaps(new []{ orderMap });
```

I find the test fixture really hard to follow. Could you make it a bit more concrete? Like maybe with dogs and cats or something.  

Stepping back a bit, what does this solve that adding the $type stuff in JSON doesn't solve?

slewis74:
Appreciate the feedback. This is something I don't particularly like either, but got me to where I could validate the rest of the approach. Re the integration tests, I was just trying to follow what was there atm, I agree it's very hard to follow and I'll try to refactor to make it easier.

The overall goal is around better support for custom types. You currently have to implement different raw things depending on where your type is going to end up. E.g. if it's a column then what you have to do (a custom ReaderWriter) is completely different to if it's in Json (a custom JsonConverter). My goal is to abstract that away and you just implement an `ICustomTypeDefinition`. Nevermore then knows what to wrap around that internally to make it work wherever and you don't have to worry about it.

Having this enables things like TinyTypes and also makes enabling modularity easier.

PaulStovell:
Nice! I like RelationalStoreConfiguration and how it unifies a few things. Will leave some more detailed comments later.

PaulStovell:
Should we rename AmazingConverter? Previously it was funny because it was internal, but if it will be public it needs a better name I think

source/Nevermore.IntegrationTests/CustomTypes/CustomSingleTypeInColumnFixture.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    CustomTypeUsedAsAProperty =  new CustomTypeUsedAsAProperty
+                    {
+                        Name = "foo",
+                        Value = 10
+                    }
+                };
+                transaction.Insert(model);
+
+                var read = transaction.Query<CustomTypeWithColumnToTestSerialization>()
```
PaulStovell, line 27: I still find this test a bit confusing - after a few minutes I think I figured out that it's testing that it can read and write it as both a string and a custom type, but I'm not sure. If the example was more real world or had comments maybe that might help?

source/Nevermore.IntegrationTests/CustomTypes/CustomSingleTypeInColumnFixture.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        class CustomTypeWithColumnToTestSerializationMap : DocumentMap<CustomTypeWithColumnToTestSerialization>
+        {
+            public CustomTypeWithColumnToTestSerializationMap()
+            {
+                TableName = "ModelTypeWithColumn";
+                
+                Column(m => m.CustomTypeUsedAsAProperty);
+                Column(m => m.JSON);
```
PaulStovell, line 110: Mapping the `JSON` column shouldn't be necessary?

source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
```diff
@@ -2,21 +2,20 @@
...
 using Nevermore.RelatedDocuments;
-using Newtonsoft.Json;
 
 namespace Nevermore.IntegrationTests
 {
     public class IntegrationTestDatabase
     {
         readonly TextWriter output = Console.Out;
-        readonly string SqlInstance = Environment.GetEnvironmentVariable("NevermoreTestServer") ?? "(local)\\SQLEXPRESS,1433";
+        readonly string SqlInstance = Environment.GetEnvironmentVariable("NevermoreTestServer") ?? "(local)";
```
PaulStovell, line 18: Not sure this should be changed

source/Nevermore.Tests/RelationalStoreFixture.cs
```diff
@@ -25,7 +25,7 @@ public void ShouldSetDefaultConnectionStringOptions()
         [Test]
         public void ShouldNotOverrideExplicitConnectionStringOptions()
         {
-            var store = new RelationalStore("Server=(local);Connection Timeout=123;ConnectRetryCount=123;ConnectRetryInterval=59;", "Nevermore test", null, null, null, null, 20);
+            var store = new RelationalStore("Server=(local);Connection Timeout=123;ConnectRetryCount=123;ConnectRetryInterval=59;", "Nevermore test", null, new RelationalStoreConfiguration(), null, 20);
```
PaulStovell, line 28: So much crap needed for these constructors! At least `RelationalStoreConfiguration` removes one parameter.

source/Nevermore/Mapping/AmazingConverter.cs
```diff
@@ -10,18 +10,24 @@ namespace Nevermore.Mapping
     /// The one and only <see cref="AmazingConverter" />. Can convert from absolutely anything to absolutely
     /// anything.
     /// </summary>
-    public static class AmazingConverter
+    public class AmazingConverter : IAmazingConverter
```
PaulStovell, line 13: Yep think we should rename this while we have the chance

source/Nevermore/Mapping/CustomTypeDefinition.cs
```diff
@@ -0,0 +1,46 @@
...
+    public abstract class CustomTypeDefinition : CustomTypeDefinitionBase
+    {
+        /// <summary>
+        /// Get the value to write to the database. Only override this in order to take over complete control of the
+        /// serialization for both columns and JSON.
+        /// </summary>
+        /// <param name="instance">The object instance being written</param>
+        /// <param name="isForJsonSerialization">True for serialization into the JSON column.</param>
+        /// <returns>The object value to write, or null to let the JsonSerializer handle the serialization.</returns>
+        public virtual object ToDbValue(object instance, bool isForJsonSerialization)
```
PaulStovell, line 14: I don't get this, why is it `ToDbValue` when it could be `isForJson`? 

I feel like I'd expect this type interface to have different methods for the different types of conversion.

PaulStovell:
Left some feedback but overall I can't shake the feeling that as a consumer of this API, it's not clear:

 - Why I should use it
 - How I should go about using it

And if I were to use it, it feels a bit complex. E.g., what would I use TypeDesignatingTypeDiscriminator for? If I had a problem I don't think I'd immediately search for a class with that name. 

I wonder what would happen if you were to write a wiki page explaining how to use it with some real-world-ish examples. It might help to make the API a bit more usable? E.g., from an empty console app with Nevermore installed, how do I put all this together. 

(I think my concern here is that we're adding a lot of complexity to Nevermore in pursuit of things that may not be worth it?)

source/Nevermore.IntegrationTests/CustomTypes/CustomSingleTypeInColumnFixture.cs
slewis74: This is a pattern that a number of the tests in here use, they map the JSON as a string and then compare to ensure it's exactly what's expected. This actually showed up some quirks in things I was testing, where the raw JSON showed some nested objects were being serialized as strings rather than objects, which all worked with new data going in/out but would fail with legacy data.

source/Nevermore.Contracts/Nevermore.Contracts.csproj
```diff
@@ -20,7 +20,7 @@
     <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">True</GeneratePackageOnBuild>
     <IncludeSymbols>true</IncludeSymbols>
     <WarningsAsErrors />
-    <TargetFramework>netstandard1.2</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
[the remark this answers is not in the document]
slewis74: I think there was an assumption that had to change at one point, but I checked and it isn't required now so I've rolled it back

source/Nevermore.IntegrationTests/IntegrationTestDatabase.cs
slewis74: 👍 reverted

Issue #96, opened 2020-04-15
Nevermore 12.0


PaulStovell:
Nevermore 12.0 is a major refactor, rewrite and design change to Nevermore. This issue summarises some of the key changes and what I was trying to do when working on it. 

## Highlights

A lot has changed! Here are the main reasons for the changes and why you should care about upgrading:

 - ⏩ **Faster!** Nevermore 12.0 is faster for most queries, and when reading larger documents, it's much faster and uses less memory (up to 75% less). Many queries are significantly optimized. 
 - 💀 **Nevermore.Contracts is dead**. You no longer need `IId` or `IDocument`. This library is gone. 
 - 🕐 **Async support**. You have been _awaiting_ a long time for this 😉  
 - 🌲 **Extensible**. You can add new types, handle inherited documents, and more, without changing existing code. 
 - 🌮 **New queries**. Documents are boring. Stream plain classes, tuples, and more.
 - ✂️ **Read/write**. Split `IReadTransaction` and `IWriteTransaction`.
 - 🤐 **Compression**. Use compression for large documents.

There are some breaking changes which will be outlined below, but the highlights are:

 - Some method signatures have changed
 - `IId`, `IDocument` etc. are gone - projects that need them will need to copy the code
 - `AmazingConverter` and some other internal classes are gone
 - `ReferenceCollection` is gone, if you need this there's a shim you can copy

⚠️  **Because so much changed, make sure you test your application thoroughly upon upgrading**

## Performance improvements

A benchmarking project has been added. Here's how old Nevermore 12.0 compares to (on `master`) on my computer running Windows (bare metal) against a local SQL Express instance:

![image](https://user-images.githubusercontent.com/47085/79324985-c3ab0680-7f53-11ea-98d4-048134d5fd72.png)

These improvements come about thanks to:

 - Caching of the various plans we make around how we read and map documents
 - Using `CommandBehavior.SequentialAccess` when querying, and changing all reading code to support it
 - When reading documents > 1KB, instead of reading them as a giant string (which for 85KB documents sends them to the LOH), we read them directly in a buffer from the data reader. This should improve load and query performance for at least 20% of the documents in a large Octopus customer database, or 25% of the documents in Octofront. 

`Load(string[])` got improved quite a bit as the new Nevermore pulls in #92.

```
Loaded 50000 products in 7347ms (old Nevermore)
Loaded 50000 products in 322ms (new Nevermore)
```

[...]

PR #97, opened 2020-04-15
Nevermore 2.0 PR

PaulStovell:
This is the PR that merges "Nevermore 2.0". 

For an overview of what this PR introduces, see #96 

This PR merges or makes obsolete the following PR's:

 - #93 - this PR adds async support
 - #92 - this PR incorporates this improvement
 - #87 - this is now possible, including removal of `IId`
 - #85 and #94 - extensibility points have been added for all of these scenarios

**Why did this PR snowball so much?** Probably I'm just really bad at managing scope creep. 😄 

The main reason is that adding `async` support meant choosing between making all the sync versions call the async versions and then `Wait()`, which might make Octopus slower, or duplicating all the methods. I decided to make the async versions standalone. 

In creating async versions, I hit a number of places that were inconsistent (e.g., you can Insert a document with a custom table name, but not update it, or provide hints for one method but not another...). Methods like `ExecuteReader` with a projection and other places did their own mapping. 

In trying to benchmark and performance tune, I then hit a few cases where we've got half-baked implementations for extensibility that weren't being used, but made it hard to cache some of the work that runs in tight loops when reading/writing documents. 

I also wanted to add support for things like tuples, plain classes and reading lists of basic types. All of these needed some kind of extensibility story (Octofront have our own - we want to store NodaTime values in columns without custom reader/writers on every column), which is why it ended up incorporating some of Shannon's ideas around relational store configuration. 

## How to review this PR

Most of the code is ready to be reviewed, although it's not ready to be merged yet until a bit more test coverage is achieved. All the existing tests pass (on my machine anyway - still fighting TeamCity). 

What started as a surgical set of changes which would have been easily reviewed became something much bigger, and because it was adding so many breaking changes anyway, it became a good time to break other things - this will probably make it hard to review the complete PR, so I don't expect anyone to. 

I'm working on a migration guide which might actually be a better way to see what changed, and then you can come and find the code in here to see how it's implemented.


PaulStovell:
This was merged to `master` (squashed)

PR #105, opened 2020-05-07
Fix default schema usage on configuration

tunger:
I noticed that when simply setting the default schema name on the configuration like this, it wouldn't work.
```csharp
var config = new RelationalStoreConfiguration(ConnectionString);
config.DefaultSchema = "TestSchema";
```

That's because the configuration was propagated down from `DocumentMapRegistry`, overwriting values on map and related map.
Instead I would vote for leaving schema name on maps as they are, and evaluate actual schema name when building the queries. That way we can use the `DefaultSchema` property on the configuration, and don't need to create a new `DocumentMapRegistry` before adding mappings.


PaulStovell:
Nice solution, thanks!

PR #107, opened 2020-05-25
Adds the ability to have a Subquery as Source

gupta-kartik:


PaulStovell:
Can you include a test that shows how the feature would be used end-to-end? The current test shows how to call AsSource but not how to combine it with anything.

source/Nevermore/Advanced/SourceQueryBuilder.cs
```diff
@@ -71,7 +71,8 @@ public class SubquerySourceBuilder<TRecord> : SourceQueryBuilder<TRecord>, ISubq
         readonly ISelect select;
         string alias;
 
-        public SubquerySourceBuilder(ISelect select,
+        public SubquerySourceBuilder(ISelect select, 
+            string alias,
```
matt-richardson, line 75: This is a tad confusing - this adds the alias as a constructor param, when there's already a builder method to set the alias. If you removed this param, that'd ditch the need to have a `null` [here](https://github.com/OctopusDeploy/Nevermore/pull/107/files?file-filters%5B%5D=.cs#diff-259ccf05853011d388caa6554900d568R183)

PR #109, opened 2020-06-10
Added `IQueryLogger` which gives the ability to hook in telemetry

droyad:
Brings #108 to v12


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -196,21 +196,19 @@ public IAsyncEnumerable<TRecord> StreamAsync<TRecord>(string query, CommandParam
...
+            using var reader = ExecuteReaderInternal(command);
+            timeTillFirstRecord = timedSection.ElapsedMilliseconds;
+
             foreach (var item in ProcessReader<TRecord>(reader, command))
                 yield return item;
         }
 
-        public async IAsyncEnumerable<TRecord> StreamAsync<TRecord>(PreparedCommand command, [EnumeratorCancellation] CancellationToken cancellationToken = default)
-        {
-            await using var reader = await ExecuteReaderAsync(command, cancellationToken);
```
droyad, line 206: The previous logging would give a false timing as the `CommandExecutor` would be disposed at the end of `ExecuteReaderAsync` before all the records had read. Essentially I think it would be measuring the time to first record and ignoring data transfer and further DB read time.

Issue #113, opened 2020-07-13
Non-string ID column not fully supported


tunger:
Quickstart states that
`The only assumption is that you'll provide an "Id" property which is a string (though it doesn't have to be a string).`

Unfortunately, there is not yet full support for IDs of other types. I tried adding support to Nevermore by changing string casts and parameters to object, but ran into problems with method overload resolution, so I'm not certain if that's the right way.

**Problem 1**
`.Load` methods only accept string. Therefore, to load entities of other types, one needs to work around by using `.Query`

**Problem 2**
`Insert` fails because Id column is cast to string.
https://github.com/OctopusDeploy/Nevermore/blob/master/source/Nevermore/Util/DataModificationQueryBuilder.cs#L246

Workaround: use a property handler for ID column that casts to string on `Read`

```csharp
    public class ReportStateCorrelationIdPropertyHandler : IPropertyHandler
    {
        public object Read(object target)
        {
            return ((ReportState)target).CorrelationId.ToString();
        }

        public void Write(object target, object value)
        {
            var state = (ReportState)target;
            Guid guid = default;
            if (value != null)
            {
                if (value.GetType() == typeof(Guid))
                    guid = (Guid)value;
                else if (value.GetType() == typeof(string))
                    guid = Guid.Parse((string)value);
            }

            state.CorrelationId = guid;
        }
    }
```

Would be great if Nevermore could handle it. I can submit WIP PR if you want, and you can point me into right direction.

Use case is using Nevermore as MassTransit Saga repository, which uses GUIDs as correlation id.

borland:
Cleaning up: Closing this issue; it looks like it was resolved with https://github.com/OctopusDeploy/Nevermore/pull/116

PR #116, opened 2020-07-15
#113 Improve support for non-string ID columns

tunger:
This is an attempt to improve support for non-string ID columns as stated in #113.

Main changes (which need to be discussed):
- Change `Load` signatures to use `object`
- **[breaking]** Seperate `Load` methods which load multiple documents to `LoadMany`. Change signature to `object[]` or `params object[]` because `IEnumerable<object>` doesn't work for any type.

Please let me know if this is something you approve, or if we should solve this another way.

### DIFF
--- a/source/Nevermore.Benchmarks/LoadManyBenchmark.cs
+++ b/source/Nevermore.Benchmarks/LoadManyBenchmark.cs
@@ -31,7 +31,7 @@ namespace Nevermore.Benchmarks
         [Benchmark]
         public List<Customer> LoadMany()
         {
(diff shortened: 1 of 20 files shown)


tunger:
Alternative would be to create overloads for common key-types like Guid, int, long, string.

tunger:
I added tons of overloads for `string`, `int`, `long` and `Guid` as overloads for the various `Load` methods, and dropped `object`. Should be easier to use now.

PaulStovell:
I think you found a really elegant solution here - nice one!

PR #121, opened 2020-11-11
Add code to handle disposal of db connections on exception during open.

andrew-moore-octo:
In some conditions Nevermore isn't closing DB connections, which can lead to exhaustion of the connection pool. We've identified some places where connections aren't disposed. This PR fixes some of those places.

Build history: https://build.octopushq.com/buildConfiguration/OctopusDeploy_LIbraries_Nevermore?mode=builds#all-projects

The crux of this PR is this flow:

1. Calling code calls `BeginReadTransaction` or `BeginWriteTransaction`
2. A new `ReadTransaction` or `WriteTransaction` is created
3. The new transaction is added to the `RelationalTransactionRegistry` during the constructor
4. An exception is thrown during the call to `ReadTransaction.Open()` or `ReadTransaction.OpenAsync()`

Before this PR, the transaction would remain in the registry. After this PR, the transaction will be removed and disposed.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -114,6 +114,10 @@ Task("Publish")
 		Source = "https://f.feedz.io/octopus-deploy/dependencies/nuget/index.json",
 		ApiKey = EnvironmentVariable("FeedzIoApiKey")
 	});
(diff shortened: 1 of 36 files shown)


andrew-moore-octo:
[Nevermore potentially not closing some transactions](https://trello.com/c/rMyquoQk/3924-nevermore-potentially-not-closing-some-transactions)

PR #118, opened 2020-07-19
Add Nevermore.Extensions.DependencyInjection

tunger:
This adds `AddNevermore` extensions methods on `IServiceCollection`.

PR #123, opened 2020-11-13
Remove project from build.

andrew-moore-octo:
The `Nevermore.Extensions.DependencyInjection` project introduced in #118 is breaking the CI builds because the cake file expects the project to be packaged into a `nupkg`, but it's not being packaged. 

As a tactical fix to unblock the CI builds, this PR changes the `build.cake` file so that it no longer attempts to upload the `Nevermore.Extensions.DependencyInjection` nupkg file.


matt-richardson (approved):
Looks good. Nice and tactical.
Let's make sure we follow up on this and don't let it drop by the wayside.

PR #125, opened 2021-01-06
Add overloaded methods for Delete and DeleteAsync

andyinaus:
# Background
Upgrade Nevermore version to the latest in `Octopus.Server`

[Octopus Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/7965)

## Issue(s) discovered
1. `LoadMany(IEnumerable<string> ids...)` failed when ids are of different lengths
**Solution:** Instead of using `SqlMetaData.InferFromValue`, we are using our own implementation to figure out the correct `SqlMetaData` to use
2. `Delete` and `DeleteAsync` support deletion by `object` id while we only support `string`, `int`, `long` and `Guid` types for ID column
**Solution:** Add the same overload methods for `Load` and `LoadAsync` to `Delete` and `DeleteAsync`, so we only support deletion by the supported ID types

# How to review this PR

Quality :heavy_check_mark:
Correctness ✔️ 

# Pre-requisites
- [x] I have considered informing or consulting the right people, e.g.: 
  - For changes to the execution pipeline: [`#topic-execution`](https://octopusdeploy.slack.com/archives/C3KT1DFSM)
  - For other changes, consider the other `#topic-...` channels
- [x] I have considered whether this should be a patch or wait for a minor.
- [x] I have considered appropriate testing for my change.

### DIFF
--- a/source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
+++ b/source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Text;
-using Nevermore.Advanced.Serialization;
 using Nevermore.IntegrationTests.Chaos;
 using Nevermore.IntegrationTests.Contracts;
 using Nevermore.IntegrationTests.Model;
@@ -27,7 +26,10 @@ namespace Nevermore.IntegrationTests.SetUp
                 new LineItemMap(),
                 new MachineMap(),
                 new OrderMap(),
-                new MessageMap());
+                new MessageWithStringIdMap(),
+                new MessageWithIntIdMap(),
+                new MessageWithLongIdMap(),
+                new MessageWithGuidIdMap());
             
             config.TypeHandlers.Register(new ReferenceCollectionTypeHandler());
             config.InstanceTypeResolvers.Register(new ProductTypeResolver());
@@ -45,7 +47,10 @@ namespace Nevermore.IntegrationTests.SetUp
                 new LineItemMap(), 
                 new BrandMap(), 
                 new MachineMap(),
-                new MessageMap());
+                new MessageWithStringIdMap(),
+                new MessageWithIntIdMap(),
+                new MessageWithLongIdMap(),
+                new MessageWithGuidIdMap());
             
             Store = new RelationalStore(config);
         }
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -529,7 +529,7 @@ namespace Nevermore.Advanced
             var mapping = configuration.DocumentMaps.Resolve(typeof(TDocument));
 
             if (mapping.IdColumn.Type != typeof(TKey))
-                throw new ArgumentException($"Provided Id of type '{id.GetType().FullName}' does not match configured type of '{mapping.IdColumn.Type.FullName}");
+                throw new ArgumentException($"Provided Id of type '{id.GetType().FullName}' does not match configured type of '{mapping.IdColumn.Type.FullName}'.");
 
             var tableName = mapping.TableName;
             var args = new CommandParameterValues {{"Id", id}};
@@ -541,7 +541,7 @@ namespace Nevermore.Advanced
             var mapping = configuration.DocumentMaps.Resolve(typeof(TDocument));
 
             if (mapping.IdColumn.Type != typeof(TKey))
-                throw new ArgumentException($"Provided Id of type '{typeof(TKey).FullName}' does not match configured type of '{mapping.IdColumn.Type.FullName}");
+                throw new ArgumentException($"Provided Id of type '{typeof(TKey).FullName}' does not match configured type of '{mapping.IdColumn.Type.FullName}'.");
 
             var param = new CommandParameterValues();
             param.AddTable("criteriaTable", idList.ToList());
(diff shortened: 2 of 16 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -538,11 +539,13 @@ public async Task<DbDataReader> ExecuteReaderAsync(PreparedCommand preparedComma
         PreparedCommand PrepareLoadMany<TDocument, TKey>(IEnumerable<TKey> idList)
         {
             var mapping = configuration.DocumentMaps.Resolve(typeof(TDocument));
-            var tableName = mapping.TableName;
+
+            if (mapping.IdColumn.Type != typeof(TKey))
```
tothegills, line 543: Thanks for adding this check, should we include a test for it?

source/Nevermore/CommandParameterValues.cs
```diff
@@ -56,10 +56,10 @@ public CommandParameterValues(object args)
 
         public CommandType CommandType { get; set; }
 
-        public void AddTable<T>(string name, IEnumerable<T> ids)
+        public void AddTable<T>(string name, IReadOnlyCollection<T> ids)
         {
-            var idColumnMetadata = SqlMetaData.InferFromValue(ids.First(), "ParameterValue");
-
+            var idColumnMetadata = GetSqlMetaData(ids, "ParameterValue");
+
```
tothegills, line 62: Looks like some extra space here

source/Nevermore/CommandParameterValues.cs
```diff
@@ -74,7 +74,29 @@ public void AddTable(string name, TableValuedParameter tvp)
...
         }
-        
+
+        static SqlMetaData GetSqlMetaData<T>(IReadOnlyCollection<T> ids, string name)
+        {
+            var valueType = typeof(T);
+            var notSupportedErrorMsg = $"'{valueType.Name}' is not a valid ID type, supported types are: {nameof(String)}, {nameof(Int32)}, {nameof(Int64)} and {nameof(Guid)}.";
+            switch (Type.GetTypeCode(valueType))
+            {
+                case TypeCode.String: return new SqlMetaData(name, SqlDbType.NVarChar, 300);
```
tothegills, line 84: Why did you select 300 instead of buckets as Paul suggested?

source/Nevermore/IWriteQueryExecutor.cs
```diff
@@ -92,6 +92,38 @@ public interface IWriteQueryExecutor : IReadQueryExecutor
         /// <param name="cancellationToken">Token to use to cancel the command.</param>
         Task UpdateAsync<TDocument>(TDocument document, UpdateOptions options, CancellationToken cancellationToken = default) where TDocument : class;
 
+        /// <summary>
```
tothegills, line 95: I think this more explicit interface is nice. This is breaking change to the public API so should be versioned appropriately.

source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
```diff
@@ -27,7 +26,10 @@ protected FixtureWithRelationalStore()
                 new LineItemMap(),
                 new MachineMap(),
                 new OrderMap(),
-                new MessageMap());
+                new MessageWithStringIdMap(),
```
tothegills, line 29: 👍

tothegills (commented):
Thanks for submitting this PR, I think the delete interface in particular is much nicer.
I'm curious about keeping the constant column length for strings vs using a bucket? Paul's advise was that a bucket would allow users of Nevermore longer column lengths without needing to modify Nevemore.

source/Nevermore/CommandParameterValues.cs
andyinaus: `GetBestSizeBucket()`:
1. will use the wrong size if the length is longer than 200
2. the `ParameterValue` column can only be of length 300 at most now, using this method may give people the impression that we support any length
3. this suggestion was to help us optimize the execution plan if we wish to derive the length of the column from the id with the maximum length instead of specifying a fixed-length like 300

image

source/Nevermore/Advanced/ReadTransaction.cs
andyinaus: Sure, I will add all tests for this kind of check in other places where we use `PrepareXXXXXX` command.

source/Nevermore/CommandParameterValues.cs
tothegills: 1: We could introduce a new 300 bucket. It does seem arbitrary. We'd need `ParamaterValue` to align with a bucket size?
2: `ParameterValue` can be adjusted by modifying the table, so Nevermore would support any value but provides a default of 300.
3: A fixed or bucket column length results in predictable query plans. The bucket suggestion was to keep the query plan optimisation while also allowing consumers of Nevermore to use a longer `ParameterValue` if they required it.

To me using the bucket seems like a more flexible option while retaining performance.

andyinaus: We agreed to use bucket size here to gain flexibility and add some comment to explain our decision in code.

andyinaus: hmm ... looks like it truncates the string value to an empty string if the `SqlMetaData` has a max length of 0.

image

Let's catch up tmr morning before our daily sync.

andyinaus: To Summarize:

We agreed to use a fixed length of 300 for `SqlMetaData` used for the `ParameterValue` column for now, which also aligns with the column length in our table schema in `Script0002-ParameterList.sql` due to time constraint.

We would probably revisit this issue later (perhaps during sharpening) if there is a need to dynamically resolve the max length based on each DocumentMap configuration.

tothegills (approved):
👍

PR #127, opened 2021-03-09
Fixed properties serialising when mapped to columns

YuKitsune:
## Summary

In scenarios where derived classes have their own tables, the serialised JSON would include properties that were already mapped to columns. This is because the `RelationalJsonContractResolver` used the `DeclaringType` to determine what `Type` of model is being resolved.
The issue with this is that `DeclaringType` will get the `Type` that the property was declared on. In the below example, the `Registration` property is declared on `Vehicle`, so the `RelationalJsonContractResolver` would look for a mapping for the `Vehicle` class when the mapping is actually be defined for the `Car` class.

```csharp
public abstract class Vehicle 
{
  public string Registration { get; set; }
}

public class Car : Vehicle
{
}
```

## Ideal Solution
The ideal solution would be to change `DeclaringType` to `ReflectingType`, which should return the `Type` for `Car`, but this won't work as Netwonsoft.Json actually gets the `MemberInfo` from the declaring type (vehicle), so that it can access any potential private getters/setters. (Issue open on the Newtonsoft.Json repo for this https://github.com/JamesNK/Newtonsoft.Json/issues/2488)

## Chosen Solution
This PR addresses this issue by overriding the `CreateProperties` method (since the target `Type` is exposed as a parameter), then passing the target `Type` as a parameter to a private `CreateProperty` method.

### DIFF
--- a/source/Nevermore/RelationalJsonContractResolver.cs
+++ b/source/Nevermore/RelationalJsonContractResolver.cs
@@ -1,6 +1,6 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Reflection;
-using Nevermore.Mapping;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 using System.Linq;
@@ -16,9 +16,20 @@ namespace Nevermore
             this.configuration = configuration;
         }
 
-        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+        protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
         {
-            configuration.DocumentMaps.ResolveOptional(member.DeclaringType, out var map);
+            var members = GetSerializableMembers(type);
+
+            return members.Select(member => CreateProperty(type, member, memberSerialization)).ToList();
+        }
+
+        JsonProperty CreateProperty(Type type, MemberInfo member, MemberSerialization memberSerialization)
+        {
+            // Todo: Newtonsoft derives the MemberInfo from the base class if that's where the member is defined,
+            //  so ReflectedType will always be the same as DeclaredType. Issue currently open to fix this:
+            //  https://github.com/JamesNK/Newtonsoft.Json/issues/2488
+            // configuration.DocumentMaps.ResolveOptional(member.ReflectedType, out var map);
+            configuration.DocumentMaps.ResolveOptional(type, out var map);
 
             var property = base.CreateProperty(member, memberSerialization);
 
--- a/source/Nevermore/RelationalJsonContractResolver.cs
+++ b/source/Nevermore/RelationalJsonContractResolver.cs
@@ -19,8 +19,25 @@ namespace Nevermore
         protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
         {
             var members = GetSerializableMembers(type);
+            if (members == null)
+            {
+                throw new JsonSerializationException("Null collection of serializable members returned.");
+            }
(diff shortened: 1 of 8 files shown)


YuKitsune:
[Fix Nevermore bug to do with not removing JSON properties that get written to a column](https://trello.com/c/1xyjD0fz/117-fix-nevermore-bug-to-do-with-not-removing-json-properties-that-get-written-to-a-column)

flin-8:
Have you tried creating an Octopus Server PR that uses this build? To see if anything has broken/changed because of this?

source/Nevermore/RelationalJsonContractResolver.cs
```diff
@@ -16,9 +16,20 @@ public RelationalJsonContractResolver(IRelationalStoreConfiguration configuratio
             this.configuration = configuration;
         }
 
-        protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+        protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
```
flin-8, line 19: This is the base implementation, should we copy that?

image

YuKitsune: I did try that, but ran into an issue with the `GetNameTable` method. It's marked as `internal`, and returns the value of a private field. It looks like it's referenced a fair bit in the `JsonTextReader`, but all usages I could find have alternatives if the `JsonNameTable` doesn't yield any result.
I think it would be fine to exclude the name table related parts of the method if we intend to copy it. Thoughts?

flin-8: Calling JSON expert @slewis74

YuKitsune:
> Have you tried creating an Octopus Server PR that uses this build? To see if anything has broken/changed because of this?

Octopus branch here with the changes: https://github.com/OctopusDeploy/OctopusDeploy/pull/8357
Tested with an existing and a fresh instance and it seemed to work as expected

source/Nevermore/RelationalJsonContractResolver.cs
YuKitsune: Looking through the usages, the `JsonNameTable` appears to be a form of cache for property names. I would be confident in implementing this:
Screen Shot 2021-03-11 at 10 14 52 am
It's as close to the original as we can make it

source/Nevermore/RelationalJsonContractResolver.cs
```diff
@@ -16,9 +16,37 @@ public RelationalJsonContractResolver(IRelationalStoreConfiguration configuratio
...
+                var property = CreateProperty(type, member, memberSerialization);
+
+                if (property != null)
+                {
+                    properties.AddProperty(property);
+                }
+            }
+
+            var orderedProperties = properties.OrderBy(p => p.Order ?? -1).ToList();
+            return orderedProperties;
```
YuKitsune, line 40: This is mostly copied from the base class with the exception of the `DefaultJsonNameTable`, since that's an internal field. It seems like it's used as a cache, and the tests are all passing with this change, so I'm fairly confident this is solid 👍 

https://github.com/JamesNK/Newtonsoft.Json/blob/f7e7bd05d9280f17993500085202ff4ea150564a/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L1362-L1390

APErebus (approved):
Looks good! Appreciate the test coverage as well.

MarkSiedle:
This never got actioned. Closing for clarity (please re-open if this is still in-progress).

PR #128, opened 2021-04-12
Added Row Versioning

pawelpabich:
# Problem #

It is possible for N > 1 independent operations to try to modify the same data at more or less the same time. 
This can lead to operations overriding each other data.

This PR adds support for [row versioning](https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15) which will ensure that the data in the database hasn't changed since it was retrieved and the update is safe to be applied.

# Why now #

Octopus Cloud is using background tasks to managed Cloud instances.  At the moment it is possible to for N > 1 tasks to modify the data of the same instance. To prevent this from happening we are adding 2 safety nets:
1. [Tasks will need to acquire a lock before they can modify a cloud instance](https://github.com/OctopusDeploy/Octofront/pull/2636)
2. Row versioning will prevent tasks from overwriting each other data in cases when they didn't acquire a lock 

# Other #

The new code introduces additional overhead only when row versioning is enabled.

### DIFF
--- a/source/Nevermore.IntegrationTests/Advanced/HooksFixture.cs
+++ b/source/Nevermore.IntegrationTests/Advanced/HooksFixture.cs
@@ -22,13 +22,15 @@ namespace Nevermore.IntegrationTests.Advanced
             var customer = new Customer {Id = "C131", FirstName = "Fred", LastName = "Freddy"};
             transaction.Insert(customer);
             AssertLogged(log, "BeforeInsert", "AfterInsert");
-            
+
+            customer = transaction.Load<Customer>("C131");
+
             transaction.Update(customer);
             AssertLogged(log, "BeforeUpdate", "AfterUpdate");
 
             transaction.Delete(customer);
             AssertLogged(log, "BeforeDelete", "AfterDelete");
-            
+
             transaction.Commit();
             AssertLogged(log, "BeforeCommit", "AfterCommit");
         }
@@ -62,7 +64,7 @@ namespace Nevermore.IntegrationTests.Advanced
             {
                 this.log = log;
             }
-            
+
             public void BeforeInsert<TDocument>(TDocument document, DocumentMap map, IWriteTransaction transaction) where TDocument : class => log.AppendLine(nameof(BeforeInsert));
             public void AfterInsert<TDocument>(TDocument document, DocumentMap map, IWriteTransaction transaction) where TDocument : class => log.AppendLine(nameof(AfterInsert));
             public void BeforeUpdate<TDocument>(TDocument document, DocumentMap map, IWriteTransaction transaction) where TDocument : class => log.AppendLine(nameof(BeforeUpdate));
--- a/source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
+++ b/source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
@@ -1,0 +1,78 @@
+using System;
+using System.IO;
+using System.Linq;
+using FluentAssertions;
+using Nevermore.IntegrationTests.Model;
+using Nevermore.IntegrationTests.SetUp;
+using NUnit.Framework;
+
+namespace Nevermore.IntegrationTests.Advanced
+{
+    public class RowVersionFixture : FixtureWithRelationalStore
+    {
+        [Test]
+        public void UpdateChangesRowVersion()
+        {
+            using var transaction = Store.BeginTransaction();
+            transaction.Insert( new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"});
+            transaction.Commit();
+
+            var customer = transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single();
+            customer.LastName = "LastName2";
+            transaction.Update(customer);
+
+            var updatedCustomer = transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single();
+
+            customer.RowVersion.Should().NotEqual(updatedCustomer.RowVersion);
+        }
+
+        [Test]
+        public void DetectsPartiallyPopulatedDocuments()
+        {
+            var customer = new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"};
+            RunInTransaction(transaction => transaction.Insert(customer));
+
+            customer.FirstName = "FirstName1";
+            Action invalidUpdate = () => RunInTransaction(transaction => transaction.Update(customer));
+            invalidUpdate.ShouldThrow<InvalidDataException>();
+        }
+
+        [Test]
+        public void FailsUpdateWhenDataBecomesStale()
+        {
+            RunInTransaction(transaction =>
+            {
+                transaction.Insert( new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"});
+            });
+
+            var customer1 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+            var customer2 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+
+            customer1.LastName = "LastName1";
+            RunInTransaction(transaction => transaction.Update(customer1));
+
+            customer2.LastName = "LastName2";
+            Action invalidUpdate = () => RunInTransaction(transaction => transaction.Update(customer2));
+
+            invalidUpdate.ShouldThrow<StaleDataException>();
+        }
+
+        TResult RunInTransaction<TResult>(Func<IRelationalTransaction, TResult> func)
+        {
+            using var transaction = Store.BeginTransaction();
+            var result =  func(transaction);
+            transaction.Commit();
+
+            return result;
+        }
+
+        void RunInTransaction(Action<IRelationalTransaction> action)
+        {
+            RunInTransaction(transaction =>
+            {
+                action(transaction);
+                return string.Empty;
+            });
+        }
+    }
+}
--- a/source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
+++ b/source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
@@ -1,5 +1,4 @@
 using System;
-using System.Linq;
 using FluentAssertions;
 using Nevermore.IntegrationTests.Model;
 using Nevermore.IntegrationTests.SetUp;
@@ -15,7 +14,7 @@ namespace Nevermore.IntegrationTests.Advanced
             var customer1 = new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"};
             RunInTransaction(transaction => transaction.Insert(customer1));
 
-            var customer2 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+            var customer2  = RunInTransaction(transaction => transaction.Load<Customer>(customer1.Id));
 
             customer2.RowVersion.Should().Equal(customer1.RowVersion);
 
@@ -36,23 +35,27 @@ namespace Nevermore.IntegrationTests.Advanced
 
             customer = RunInTransaction(transaction => transaction.Load<Customer>(customer.Id));
 
+            var previousRowVersion = customer.RowVersion;
             customer.LastName = "LastName1";
             RunInTransaction(transaction => transaction.Update(customer));
 
+            customer.RowVersion.Should().NotEqual(previousRowVersion);
+
+            previousRowVersion = customer.RowVersion;
             customer.LastName = "LastName2";
             RunInTransaction(transaction => transaction.Update(customer));
+
+            customer.RowVersion.Should().NotEqual(previousRowVersion);
         }
 
         [Test]
         public void FailsUpdateWhenDataBecomesStale()
         {
-            RunInTransaction(transaction =>
-            {
-                transaction.Insert( new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"});
-            });
+            var customer = new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"};
+            RunInTransaction(transaction => transaction.Insert( customer));
 
-            var customer1 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
-            var customer2 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+            var customer1 = RunInTransaction(transaction => transaction.Load<Customer>(customer.Id));
+            var customer2 = RunInTransaction(transaction => transaction.Load<Customer>(customer.Id));
 
             customer1.LastName = "LastName1";
             RunInTransaction(transaction => transaction.Update(customer1));
@@ -61,6 +64,9 @@ namespace Nevermore.IntegrationTests.Advanced
             Action invalidUpdate = () => RunInTransaction(transaction => transaction.Update(customer2));
 
             invalidUpdate.ShouldThrow<StaleDataException>();
+
+            var customer3 = RunInTransaction(transaction => transaction.Load<Customer>(customer.Id));
+            customer3.LastName.Should().Be(customer1.LastName);
         }
 
         TResult RunInTransaction<TResult>(Func<IRelationalTransaction, TResult> func)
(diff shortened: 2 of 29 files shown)


source/Nevermore.IntegrationTests/Advanced/HooksFixture.cs
```diff
@@ -22,13 +22,15 @@ public void ShouldCallHooks()
             var customer = new Customer {Id = "C131", FirstName = "Fred", LastName = "Freddy"};
             transaction.Insert(customer);
             AssertLogged(log, "BeforeInsert", "AfterInsert");
-            
+
+            customer = transaction.Load<Customer>("C131");
```
flin-8, line 26: So once I insert, I need to reload to make an update? Could Nevermore fill in the `RowVersion` for me?

source/Nevermore/Mapping/ColumnMapping.cs
```diff
@@ -34,9 +35,10 @@ internal ColumnMapping(string columnName, Type type, IPropertyHandler handler, P
         public Type Type { get; }
         public IPropertyHandler PropertyHandler { get; private set; }
         public PropertyInfo Property { get; }
-        
+
         public int? MaxLength => maxLength;
         public ColumnDirection Direction => direction;
+        public bool RowVersion => rowVersion;
```
flin-8, line 41: `RowVersion` -> `IsRowVersion`

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -75,17 +76,22 @@ public PreparedCommand PrepareUpdate(object document, UpdateOptions options = nu
                 default:
                     throw new ArgumentOutOfRangeException();
             }
-            
+
+            var whereStatements = new List<string> {$"[{mapping.IdColumn.ColumnName}] = @{mapping.IdColumn.ColumnName}"};
+            whereStatements.AddRange(mapping.RowVersionColumns().Select(dvc => $"[{dvc.ColumnName}] = @{dvc.ColumnName}"));
```
flin-8, line 81: Can you have more than 1 column of type `RowVersion`?

source/Nevermore/Mapping/ColumnMapping.cs
andyinaus: `IsRowVersioned` may be better

source/Nevermore.IntegrationTests/Advanced/HooksFixture.cs
pawelpabich: 👍

source/Nevermore/Mapping/ColumnMapping.cs
pawelpabich: This is gone now

source/Nevermore/Util/DataModificationQueryBuilder.cs
pawelpabich: I've seen examples where more than one column was used but we can be opinionated here.

pawelpabich:
@flin-8 @andyinaus . What do you think about this approach ? https://github.com/OctopusDeploy/Nevermore/pull/128/commits/2b4d900cca394d37af3851fbe95556394884e6b0

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -74,25 +83,60 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
...
+            var numberOfUpdatedRows = ExecuteNonQuery(command);
+
+            AssertUpdateHasBeenSuccessful(numberOfUpdatedRows, command);
+            RefreshRowVersionIfRequired(document, command);
+
             configuration.Hooks.AfterUpdate(document, command.Mapping, this);
             configuration.RelatedDocumentStore.PopulateRelatedDocuments(this, document);
         }
 
+        void RefreshRowVersionIfRequired<TDocument>(TDocument document, PreparedCommand command) where TDocument : class
```
flin-8, line 114: Not sure if you want to make this optimization:
https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15

You can get the new RowVersion from the same `UPDATE` query (last code snippet)

flin-8:
> @flin-8 @andyinaus . What do you think about this approach ? [2b4d900](https://github.com/OctopusDeploy/Nevermore/commit/2b4d900cca394d37af3851fbe95556394884e6b0)

I like it! Declaring a specific `RowVersion` column is more intuitive.

Also, I'm pretty sure you can only have 1 per table anyway

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -217,5 +265,11 @@ public async Task CommitAsync(CancellationToken cancellationToken = default)
             await Transaction.CommitAsync(cancellationToken);
             await configuration.Hooks.AfterCommitAsync(this);
         }
+
+        static void AssertUpdateHasBeenSuccessful(int numberOfUpdatedRows, PreparedCommand command)
+        {
+            if (numberOfUpdatedRows == 0 && command.Mapping.RowVersionColumn != null) throw new StaleDataException("Update failed because submitted data was out of date. Refresh the data and try again.");
```
flin-8, line 271: Since we're checking the rows updated, do we want to throw a new exception if there's no `RowVersion` column but no row was updated?

source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            var customer = new Customer {FirstName = "FirstName", LastName = "LastName", Nickname = "NickName"};
+            RunInTransaction(transaction => transaction.Insert(customer));
+
+            customer.RowVersion.Should().NotBeNull();
+
+            customer = RunInTransaction(transaction => transaction.Load<Customer>(customer.Id));
+
+            customer.LastName = "LastName1";
+            RunInTransaction(transaction => transaction.Update(customer));
```
flin-8, line 40: Are you meant to assert that the row version changes after each update?

source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
```diff
@@ -0,0 +1,84 @@
...
+            var customer1 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+            var customer2 = RunInTransaction(transaction => transaction.Query<Customer>().Where(c => c.FirstName == "FirstName").ToArray().Single());
+
+            customer1.LastName = "LastName1";
+            RunInTransaction(transaction => transaction.Update(customer1));
+
+            customer2.LastName = "LastName2";
+            Action invalidUpdate = () => RunInTransaction(transaction => transaction.Update(customer2));
+
+            invalidUpdate.ShouldThrow<StaleDataException>();
```
flin-8, line 63: Also assert that the record in the DB didn't change

source/Nevermore.IntegrationTests/Model/CustomerMap.cs
```diff
@@ -11,7 +11,7 @@ public CustomerMap()
             Column(m => m.LastName).MaxLength(50);
             Column(m => m.Nickname);
             Column(m => m.Roles);
-            Column(m => m.RowVersion).LoadOnly();
+            RowVersion(m => m.RowVersion);
```
flin-8, line 14: 👍

source/Nevermore/Advanced/WriteTransaction.cs
pawelpabich: This would be a breaking change as Nevermore doesn't do that currently.

flin-8 (approved):
Overall approach looks good to me, just an optimization you might want to make when returning the rowversion and tests

source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
pawelpabich: 👍

source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
pawelpabich: Lack of exception is meant to represent that but it is a good idea to be more explicit here.

source/Nevermore/Advanced/WriteTransaction.cs
pawelpabich: That would be nice. I don't see an easy way of incorporating this into NV right now. Any ideas?

pawelpabich:
@flin-8 what do you think about this approach? It writes and reads data in the same transaction.

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -217,5 +236,77 @@ public async Task CommitAsync(CancellationToken cancellationToken = default)
...
+        async Task<object[]> ExecuteDataModificationAsync(PreparedCommand command, CancellationToken cancellationToken)
+        {
+            if (!command.Mapping.IsRowVersioningEnabled)
+            {
+                await ExecuteNonQueryAsync(command, cancellationToken);
+                return Array.Empty<object>();
+            }
+
+            var newRowVersions = new List<object>();
+            using (var reader = await ExecuteReaderAsync(command, cancellationToken))
```
pawelpabich, line 275: To make RowVersioning efficient I need to be able to read data from an insert or update.  Unfortunately, Chaos Monkey causes its regular trouble and tests fail randomly because INSERTs and UPDATEs are not retried. 

@hnrkndrssn you seemed to add that code, any ideas? 

Sample build: https://build.octopushq.com/buildConfiguration/OctopusDeploy_LIbraries_Nevermore/2282122

andyinaus:
> @flin-8 @andyinaus . What do you think about this approach ? [2b4d900](https://github.com/OctopusDeploy/Nevermore/commit/2b4d900cca394d37af3851fbe95556394884e6b0)

It may somehow indicate that `RowVersion(d=> d.RowVersion)` is not a column to me.
And this is more like a data layer concern which is very similar to [etag concept in a document database world](https://chapsas.com/understanding-optimistic-concurrency-in-cosmos-db/), so would it be better if we could keep the `RowVersion` property outside our models?

e.g.

Invoking `EnableRowVersion()` in any `<Document>Map` will add a `RowVersion` column to the corresponding DB table. We will also need to add support to retrieve the row version for a document/DB record.

source/Nevermore/Advanced/WriteTransaction.cs
hnrkndrssn: do you need to use `ExecuteReaderWithRetryAsync` maybe?

flin-8:
> It may somehow indicate that `RowVersion(d=> d.RowVersion)` is not a column to me.

`RowVersionColumn(d => d.RowVersion)` perhaps

> And this is more like a data layer concern which is very similar to [etag concept in a document database world](https://chapsas.com/understanding-optimistic-concurrency-in-cosmos-db/), so would it be better if we could keep the `RowVersion` property outside our models?

I don't think this is a data layer concern - `RowVersion` needs to travel to the client for the optimistic locking to work.

source/Nevermore/Advanced/WriteTransaction.cs
pawelpabich: The thing this is an insert that returns data. I should not be retried.  Are we getting a lot of value from Chaos monkey? Can its job achieve differently? Happy to zoom if that's easier.

andyinaus:
> > It may somehow indicate that `RowVersion(d=> d.RowVersion)` is not a column to me.
> 
> `RowVersionColumn(d => d.RowVersion)` perhaps
> 
> > And this is more like a data layer concern which is very similar to [etag concept in a document database world](https://chapsas.com/understanding-optimistic-concurrency-in-cosmos-db/), so would it be better if we could keep the `RowVersion` property outside our models?
> 
> I don't think this is a data layer concern - `RowVersion` needs to travel to the client for the optimistic locking to work.

hmm... I thought we only need to have this info cached in the memory when we, for instance, load a document back from DB. A `Document<T>` which containing the low-level DB properties such as `RowVersion` may help us, rather than making `T` itself knows the `RowVersion` property. Or we could even have a cached lookup for every document we load stored in a single transaction, and only keep low-level DB properties in this lookup and retrieve these property values for update actions when we commit the transaction.

pawelpabich:
> hmm... I thought we only need to have this info cached in the memory when we, for instance, load a document back from DB. A Document<T> which containing the low-level DB properties such as RowVersion may help us, rather than making T itself knows the RowVersion property. Or we could even have a cached lookup for every document we load stored in a single transaction, and only keep low-level DB properties in this lookup and retrieve these property values for update actions when we commit the transaction.

and 

> Invoking EnableRowVersion() in any <Document>Map will add a RowVersion column to the corresponding DB table automatically behind the scenes. We will also need to add support to retrieve the row version for a document/DB record.

Nevermore is a very thin wrapper around a SQL database.  The schema is managed by the application so we can't just add one more column. The account that is used for the app to connect to the database might not even have enough permissions to modify schema. 

Even EF, which operates at a higher level of abstraction, doesn't hide it.

PaulStovell:
Definitely feels like a concept that should exist as a core thing in Nevermore. Feel free to merge whenever.

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -217,5 +236,60 @@ public async Task CommitAsync(CancellationToken cancellationToken = default)
...
+            {
+                ApplyNewRowVersionIfRequired(documentList[i], mapping, newRowVersions[i]);
+            }
+        }
+
+        void ApplyNewRowVersionIfRequired<TDocument>(TDocument document, DocumentMap mapping, object newRowVersion) where TDocument : class
+        {
+            if (!mapping.IsRowVersioningEnabled) return;
+
+            if (newRowVersion == null) throw new StaleDataException($"Modification failed for '{typeof(TDocument).Name}' document with '{mapping.GetId(document)}' Id because submitted data was out of date. Refresh the document and try again.");
```
andyinaus, line 289: hmm ... reusing these two private methods here in both `INSERT` and `UPDATE` operations somehow makes me feel it's less readable since by just reading through the code line by line, it shows that `INSERT` operation could throw this exception as well, however, in theory, this exception will only be thrown for `UPDATE`.

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -31,40 +30,49 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
...
         {
-            var command = builder.PrepareInsert(new[] {instance}, options);
-            configuration.Hooks.BeforeInsert(instance, command.Mapping, this);
-            ExecuteNonQuery(command);
-            configuration.Hooks.AfterInsert(instance, command.Mapping, this);
-            configuration.RelatedDocumentStore.PopulateRelatedDocuments(this, instance);
+            var command = builder.PrepareInsert(new[] {document}, options);
+            configuration.Hooks.BeforeInsert(document, command.Mapping, this);
+
+            var newRowVersions = ExecuteSingleDataModification(command);
```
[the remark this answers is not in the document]
pawelpabich: 👍

source/Nevermore.IntegrationTests/Advanced/RowVersionFixture .cs
```diff
@@ -0,0 +1,146 @@
...
+        {
+            NoMonkeyBusiness();
+
+            var document1 = new DocumentWithRowVersion {Name = "Name"};
+            RunInTransaction(transaction => transaction.Insert( document1));
+
+            var document2 = new DocumentWithRowVersion {Name = "Name"};
+            Action invalidUpdate = () => RunInTransaction(transaction => transaction.Insert(document2));
+
+            invalidUpdate.ShouldThrow<UniqueConstraintViolationException>();
```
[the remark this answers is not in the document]
pawelpabich: This test was introduced because row versions broke an existing behaviour. It's very much needed here.

source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.cs
```diff
@@ -224,7 +224,17 @@ public void UpdateWithHint()
 
             this.Assent(Format(result));
         }
-  
+
+        [Test]
+        public void UpdateWithRoWVersion()
```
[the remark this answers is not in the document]
pawelpabich: hmmm :) 

image

source/Nevermore/Advanced/WriteTransaction.cs
pawelpabich: INSERT can throw here if the new row version is not returned, e.g. a bug in the SQL generation

PR #129, opened 2021-04-16
Fixed a race condition setting the retry manager. 

droyad:
This occurs when many connections are opened right after the store is created. The RetryManager will throw if one has already been set.


slewis74 (approved):
👍 Looks good

Issue #130, opened 2021-04-19
Transactions are not disposed correctly if there is an error opening them


droyad:
If there is an exception on the `Open` call within the `RelationalStore.XXXXBeginTransaction` methods the transaction is never disposed. Since it a disposable isn't returned to the caller, it is the responsibility of those methods to dispose the transaction.

Pre Nevermore 2.0 (`v12`) this was done [here](https://github.com/OctopusDeploy/Nevermore/blob/eba309452ff3fa6bc20b08dbd760d67d02ac6788/source/Nevermore/RelationalTransaction.cs#L79-L90), so this bug only affects Nevermore 12.0+

PR #131, opened 2021-04-19
Correctly disposed transactions if there was an error opening them

droyad:
Fixes https://github.com/OctopusDeploy/Nevermore/issues/130

This was fixed in 12.1 in #124 but not merged to `13.x`. I don't [like the change](https://github.com/OctopusDeploy/Nevermore/pull/124/files#diff-cb574f356346d8539d664d75fd6253faf8ab060fb248df62fd320b0f5842214fL51) to where the transaction is added to the registry so I intend to roll that back when I merge that branch.


tothegills (approved):
👍

PR #133, opened 2021-05-07
Introduce a new analyzer to ensure locally declared transactions are disposed of

hnrkndrssn:


source/Nevermore.Analyzers/Descriptors.cs
```diff
@@ -0,0 +1,94 @@
+using Microsoft.CodeAnalysis;
+// ReSharper disable InconsistentNaming
+
+namespace Nevermore.Analyzers
+{
+    internal static class Descriptors
+    {
+        internal static readonly DiagnosticDescriptor NV0001NevermoreWhereExpressionError = Create(
```
MarkSiedle, line 8: Nice, I like the numbering scheme 👍

MarkSiedle (approved):
Good test coverage explaining what's going on here 👍 I haven't played with it locally, but LGTM code-wise

source/Nevermore.Analyzers/NevermoreDisposeLocalDeclarationTransactionAnalyzer.cs
```diff
@@ -0,0 +1,51 @@
...
+                     if (declarator.Initializer is {Value: { } syntax})
+                     {
+                         var expression = (syntax as InvocationExpressionSyntax)?.Expression ??
+                                          (syntax as AwaitExpressionSyntax)?.Expression;
+
+                         if (expression is { } &&
+                             context.SemanticModel.GetSymbolInfo(expression, context.CancellationToken).Symbol is
+                                 { } symbol &&
+                             methodsWeCareAbout.Contains(symbol.Name))
+                         {
```
droyad, line 39: `GetSymbolInfo` is expensive when I was doing some perf testing, and you can get away without it (We use it [over here](https://github.com/OctopusDeploy/RoslynAnalyzers/blob/e73d8404a9979d7f352e64f775a411379d986be1/source/RoslynAnalyzers/CaseSensitiveEnumParsingAnalyzer.cs#L46))

```suggestion
                      var expression = (syntax as AwaitExpressionSyntax)?.Expression ?? syntax;

                        if (expression is InvocationExpressionSyntax ies &&
                            ies.Expression is MemberAccessExpressionSyntax memberAccess &&
                            methodsWeCareAbout.Contains(memberAccess.Name.Identifier.Text))
                        {
```

droyad (approved):

PR #139, opened 2021-05-11
Check is bypassed if a variable is passed into the Where clause

### LINKED ISSUE #137: Analyzer: Should it ignore predicates that are variables?
E.g. the following generates a NV0001 error (In [Octopus Server code](https://github.com/OctopusDeploy/OctopusDeploy/blob/35c3af7d937d5aa062e9abca2689ed8b8610d6af/source/Octopus.Core/Extensions/NevermoreQueryBuilderExtensionMethods.cs#L14))
```
    public static T FirstOrDefault<T>(this IQueryBuilder<T> query, Expression<Func<T, bool>> predicate) where T : class
        {
            return query.Where(predicate).FirstOrDefault();
        }
```

```
  NevermoreQueryBuilderExtensionMethods.cs(14, 20): [NV0001] Nevermore LINQ support will not be able to translate this expression: Cannot translate expression argument: (predicate)
```

Should we exclude this case from the analyzer?

droyad:
Fixes #137

### DIFF
--- a/source/Nevermore.Analyzers.Tests/CodeCompiler.cs
+++ b/source/Nevermore.Analyzers.Tests/CodeCompiler.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
+using System.Linq.Expressions;
 using System.Reflection;
 using Nevermore;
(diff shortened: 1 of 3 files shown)


andrew-at-octopus (approved):
Accepted

PR #140, opened 2021-05-12
Downgraded to `netstandard2.0` so it works in Visual Studio

droyad:
`netstandard2.1` analyzers don't work right in Visual Studio so we have to go backwards. Some incompatibilitly with ImmutableCollection.


andrew-at-octopus (approved):
Compiled in VS, no issues

PR #143, opened 2021-05-24
Fix SQL generation when paging with specified columns

APErebus:
Fixes a bug with SQL generation when using specified columns.

For instance
```              
 var ids = await read.Query<Table>()
                    .Column(t => t.Id)
                    .AsType<string>()
                    .ToListAsync(skip, batchSize);
```
resulted in a SQL of 
```
SELECT *
FROM (
    SELECT [Id],
    *,
    ROW_NUMBER() OVER (ORDER BY [Id]) AS RowNum
    FROM [dbo].[Table]
) ALIAS_GENERATED_1
WHERE ([RowNum] >= @_minrow)
AND ([RowNum] <= @_maxrow)
ORDER BY [RowNum]
```
which would throw an error of
> The column 'Id' was specified multiple times for 'ALIAS_GENERATED_1'.

because we were including `*` in the subquery.

The fix is to only add the default columns if there is not already specified columns.


uglybugger (approved):
Looks sensible to me 👍

PR #144, opened 2021-05-26
Fix OverrideValueIfNotSet method and move to helpers

ankithkonda:
## Background 
`OverrideValueIfNotSet` doesn’t actually check the connection string to see if a value is set. Rather it compares a new builder's property values with that of the builder that we're wanting to override.

For example, if the connection string has boolean a property set explicitly, if the new builder has the same value set by default. `OverrideValueIfNotSet` compares properties and determines that both have the same default value and so it needs to override.

So if a consumer wants to explicitly set `Trust Server Certificate=False` in the string, OverrideValueIfNotSet treats that as being equal to the builder’s default and overrides it anyway with our override value, instead of ignoring it.

## Results

This fix updates how we validate and uses the builder's ShouldSerialize which checks if the value is in the connection string.

However we also needed to copy in the DbConnectionStringKeywords from [Ms SqlClient](https://github.com/dotnet/SqlClient/blob/5c5a15d5ac842b48c4e99bff951b026f07f1a5d3/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/Common/DbConnectionStringCommon.cs#L885) as we're unable to use the indexer to set values for certain properties, and we're also unable to get the display names through reflection. So this is our fall back.

## How To Review
There should be adequate test coverage here
Quality 👀

### DIFF
--- a/source/Nevermore/Advanced/SqlConnectionStringHelpers.cs
+++ b/source/Nevermore/Advanced/SqlConnectionStringHelpers.cs
@@ -29,52 +29,52 @@ namespace Nevermore.Advanced
         // internal static readonly DbConnectionStringKeyword NamedConnection = new DbConnectionStringKeyword("Named Connection");
 
         // SqlClient
-        internal static readonly DbConnectionStringKeyword ApplicationIntent = new DbConnectionStringKeyword("Application Intent");
-        internal static readonly DbConnectionStringKeyword ApplicationName = new DbConnectionStringKeyword("Application Name");
-        internal static readonly DbConnectionStringKeyword AsynchronousProcessing = new DbConnectionStringKeyword("Asynchronous Processing");
-        internal static readonly DbConnectionStringKeyword AttachDBFilename = new DbConnectionStringKeyword("AttachDbFilename");
-        internal static readonly DbConnectionStringKeyword CommandTimeout = new DbConnectionStringKeyword("Command Timeout");
-        internal static readonly DbConnectionStringKeyword ConnectTimeout = new DbConnectionStringKeyword("Connect Timeout");
-        internal static readonly DbConnectionStringKeyword ConnectionReset = new DbConnectionStringKeyword("Connection Reset");
-        internal static readonly DbConnectionStringKeyword ContextConnection = new DbConnectionStringKeyword("Context Connection");
-        internal static readonly DbConnectionStringKeyword CurrentLanguage = new DbConnectionStringKeyword("Current Language");
-        internal static readonly DbConnectionStringKeyword Encrypt = new DbConnectionStringKeyword("Encrypt");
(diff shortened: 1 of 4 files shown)


source/Nevermore/Advanced/SqlConnectionStringHelpers.cs
```diff
@@ -0,0 +1,80 @@
...
+    public class DbConnectionStringKeyword : TinyType<string>
+    {
+        public DbConnectionStringKeyword(string value) : base(value)
+        {
+        }
+    }
+
+    // https://github.com/dotnet/SqlClient/blob/5c5a15d5ac842b48c4e99bff951b026f07f1a5d3/src/Microsoft.Data.SqlClient/netcore/src/Microsoft/Data/Common/DbConnectionStringCommon.cs#L885
+    // Updated to use TinyTypes
+    public static class DbConnectionStringKeywords
```
droyad, line 26: Can you put the contents of this just into `DbConnectionStringKeyword`?

source/Nevermore/Advanced/SqlConnectionStringHelpers.cs
```diff
@@ -0,0 +1,80 @@
+using Microsoft.Data.SqlClient;
+using Octopus.TinyTypes;
+
+namespace Nevermore.Advanced
+{
+    public static class SqlConnectionStringHelpers
+    {
+        public static void OverrideValueIfNotSet(SqlConnectionStringBuilder connectionStringBuilder, DbConnectionStringKeyword propertyName, object overrideValue)
```
droyad, line 8: I think this would be good as an extension method instead

source/Nevermore/Advanced/SqlConnectionStringHelpers.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            if (!connectionStringBuilder.ShouldSerialize(propertyName.Value))
+            {
+                connectionStringBuilder[propertyName.Value] = overrideValue;
+            }
+        }
+    }
+
+    public class DbConnectionStringKeyword : TinyType<string>
```
droyad, line 17: 👍 

Lets not add TinyTypes to Nevermore, we explicitly worked around referencing it. Instead just throw the `Value` property onto this type instead.

source/Nevermore/Nevermore.csproj
```diff
@@ -43,6 +43,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="1.3.3" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Octopus.TinyTypes" Version="0.1.539" />
```
droyad, line 46: Lets not add this

droyad (changes requested):

droyad (approved):
image

PR #145, opened 2021-05-27
Add support for identity primary keys

APErebus:
This PR adds support for identity number (`short`/`int`/`long`) primary keys.

There is a new `Identity()` method that chains off the `Id()` method in the `DocumentMap`. This marks the Id column as an identity column.

e.g. `Id(c => c.Id).Identity()`

This works by retrieving the identity value as part of the `INSERT` and updating the `Id` on the document, very similar to how the `RowVersion` is updated.

As there are now 2 output parameters, we need to change to use a table parameter to store them in. This also handles multiple inserts in a single statement

### DIFF
--- a/source/Nevermore.IntegrationTests/Model/DocumentWithIdentityIdAndRowVersionMap.cs
+++ b/source/Nevermore.IntegrationTests/Model/DocumentWithIdentityIdAndRowVersionMap.cs
@@ -1,0 +1,14 @@
+using Nevermore.Mapping;
+
+namespace Nevermore.IntegrationTests.Model
+{
+    public class DocumentWithIdentityIdAndRowVersionMap : DocumentMap<DocumentWithIdentityIdAndRowVersion>
+    {
+        public DocumentWithIdentityIdAndRowVersionMap()
+        {
+            Id(t => t.Id).Identity();
+            Column(t => t.Name);
+            RowVersion(m => m.RowVersion);
+        }
+    }
+}
--- a/source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
+++ b/source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
@@ -15,11 +15,8 @@ namespace Nevermore.IntegrationTests.SetUp
 
         protected FixtureWithRelationalStore()
         {
-            var config = new RelationalStoreConfiguration(ConnectionString);
-            config.CommandFactory = new ChaosSqlCommandFactory(new SqlCommandFactory());
-            config.ApplicationName = "Nevermore-IntegrationTests";
-            config.DefaultSchema = "TestSchema";
-            config.DocumentMaps.Register(
+            var documentMaps = new IDocumentMap[]
+            {
                 new CustomerMap(),
                 new BrandMap(),
                 new ProductMap(),
@@ -30,7 +27,18 @@ namespace Nevermore.IntegrationTests.SetUp
                 new MessageWithIntIdMap(),
                 new MessageWithLongIdMap(),
                 new MessageWithGuidIdMap(),
-                new DocumentWithRowVersionMap());
+                new DocumentWithRowVersionMap(),
+                new DocumentWithIdentityIdMap(),
+                new DocumentWithIdentityIdAndRowVersionMap()
+            };
+
+            var config = new RelationalStoreConfiguration(ConnectionString)
+            {
+                CommandFactory = new ChaosSqlCommandFactory(new SqlCommandFactory()),
+                ApplicationName = "Nevermore-IntegrationTests",
+                DefaultSchema = "TestSchema"
+            };
+            config.DocumentMaps.Register(documentMaps);
 
             config.TypeHandlers.Register(new ReferenceCollectionTypeHandler());
             config.InstanceTypeResolvers.Register(new ProductTypeResolver());
@@ -41,18 +49,7 @@ namespace Nevermore.IntegrationTests.SetUp
                 settings.ConstructorHandling = ConstructorHandling.AllowNonPublicDefaultConstructor;
             });
 
-            GenerateSchemaAutomatically(
-                new OrderMap(),
-                new ProductMap(),
-                new CustomerMap(),
-                new LineItemMap(),
-                new BrandMap(),
-                new MachineMap(),
-                new MessageWithStringIdMap(),
-                new MessageWithIntIdMap(),
-                new MessageWithLongIdMap(),
-                new MessageWithGuidIdMap(),
-                new DocumentWithRowVersionMap());
+            GenerateSchemaAutomatically(documentMaps);
 
             Store = new RelationalStore(config);
         }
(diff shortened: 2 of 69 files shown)


source/Nevermore.IntegrationTests/Model/DocumentWithIdentityIdAndRowVersionMap.cs
```diff
@@ -0,0 +1,14 @@
+using Nevermore.Mapping;
+
+namespace Nevermore.IntegrationTests.Model
+{
+    public class DocumentWithIdentityIdAndRowVersionMap : DocumentMap<DocumentWithIdentityIdAndRowVersion>
+    {
+        public DocumentWithIdentityIdAndRowVersionMap()
+        {
+            Id(t => t.Id).Identity();
```
APErebus, line 9: This is how you define an `IDENTITY` primary key column

source/Nevermore.IntegrationTests/SetUp/FixtureWithRelationalStore.cs
```diff
@@ -95,7 +92,6 @@ void GenerateSchemaAutomatically(params IDocumentMap[] mappings)
                 {
                     SchemaGenerator.WriteTableSchema(map.Build(), null, schema);
                 }
-                schema.AppendLine($"ALTER TABLE [TestSchema].[{nameof(DocumentWithRowVersion)}] ADD [RowVersion] rowversion");
```
APErebus, line 98: Handled in SchemaGenerator now

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -530,7 +530,7 @@ protected TResult[] ReadResults<TResult>(PreparedCommand preparedCommand, Func<D
             return command.ReadResults(mapper);
         }
 
-        protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand preparedCommand, Func<DbDataReader, TResult> mapper)
+        protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand preparedCommand, Func<DbDataReader, Task<TResult>> mapper)
```
APErebus, line 533: ReadResultsAsync now is an async mapper func (async all the way down)

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -206,6 +214,12 @@ public string AllocateId(Type documentType)
             return AllocateId(mapping);
         }
 
+        public string AllocateId<TDocument>()
+        {
+            var mapping = configuration.DocumentMaps.Resolve<TDocument>();
+            return AllocateId(mapping);
+        }
```
APErebus, line 221: Handy new overload (save a lot of `AllocateId(typeof(MyDocument))`)

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -216,7 +230,7 @@ public string AllocateId(string tableName, string idPrefix)
             return AllocateId(tableName, key => $"{idPrefix}-{key}");
```
APErebus, line 230: This overload is a problem. Ideally you shouldn't be able to Allocate an Id for an Identity column (the db is doing it for you) and it should throw an exception, however you only receive a string name, so we either have to look up the mapping (which we are already doing above) or we just have to let the code fail badly.

I would be in favour of removing the public API `string AllocateId(string tableName, string idPrefix)` and only allowing public API's that take something that has a DocumentMap, where we can change this.

How often are you allocating Id's for a table that you don't have a document map for... seems weird

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -238,59 +252,118 @@ public async Task CommitAsync(CancellationToken cancellationToken = default)
             await configuration.Hooks.AfterCommitAsync(this);
         }
 
-        object[] ExecuteDataModification(PreparedCommand command)
+        DataModificationOutput[] ExecuteDataModification(PreparedCommand command)
         {
-            if (!command.Mapping.IsRowVersioningEnabled)
+            if (!command.Mapping.HasModificationOutputs)
```
APErebus, line 257: there are 2 different possible modification outputs, the RowVersion and/or the new Id

source/Nevermore/Mapping/IdColumnMapping.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Nevermore.Mapping
+{
+    public class IdColumnMapping : ColumnMapping, IIdColumnMappingBuilder
```
APErebus, line 7: This is a new inherited version of `ColumnMapping`, specific to the Id column, allowing for specific Id column mapping

source/Nevermore/Transient/DbCommandExtensions.cs
```diff
@@ -36,71 +36,57 @@ public static Task<int> ExecuteNonQueryWithRetryAsync(this DbCommand command, Re
...
-                            command.Connection.State == ConnectionState.Open)
-                            command.Connection.Close();
-                        throw;
-                    }
-                });
-            }
-            catch (Exception ex)
-            {
-                throw new Exception($"Exception occurred while executing a reader for `{command.CommandText}`", ex);
-            }
```
APErebus, line 74: This exception handler was bad:tm:

It was catching unique constraint exception too early in the call chain and hiding the underlying sql error. The reason this changed in this PR is that previously it only threw the exception when the reader made the 1st call to `.Read()`, but the exception is now thrown here because the generated SQL is different and the exception is thrown at this point, rather than when read.

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -39,6 +39,9 @@ public PreparedCommand PrepareInsert(IReadOnlyList<object> documents, InsertOpti
             options ??= InsertOptions.Default;
             var mapping = GetMapping(documents);
 
+            if (mapping.IdColumn.IsIdentity && options.CustomAssignedId != null)
+                throw new InvalidOperationException($"{nameof(InsertOptions)}.{nameof(InsertOptions.CustomAssignedId)} is not supported for identity Id columns.");
```
APErebus, line 43: You obv can't set an custom assigned Id with an identity column

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -457,7 +491,19 @@ class RelatedDocumentTableData
 
     internal static class DataModificationQueryBuilderExtensions
     {
+        static readonly Dictionary<Type, string> IdentityIdTypeMap = new Dictionary<Type, string>
+        {
+            [typeof(short)] = "smallint",
+            [typeof(int)] = "int",
+            [typeof(long)] = "bigint"
```
APErebus, line 498: because we have to define a `TABLE` row version, we need the sql type that matches our `Id` column.

source/Nevermore/Advanced/WriteTransaction.cs
slewis74: Octopus server does exactly this to allocate Tag Ids. They are technically a child of TagSet, but the Ids are issued as though it was a table of its own. Agreed it is an edge case usage, but it is used.

slewis74 (approved):
Once #149 merges in to here I think this is going to be 👍

slewis74:
[This page](https://github.com/OctopusDeploy/Nevermore/wiki/Documents) in the Wiki should be updated after this PR merges

PR #147, opened 2021-06-11
Extending cancellationToken's use in async calls

andrew-at-octopus:
Found that there were a few async operations that did not use/accept a cancellation token:

- ExecuteNonQueryWithRetryAsync method in Nevermore.Transient.DbCommandExtensions - meaning it waited for timeout or completion
- OpenWithRetryAsync - meaning it would timeout rather than cancelled
- ReadResultsAsync - didn't accept a cancellation token, so you couldn't interrupt the reading of the resultset

Also found that on a case-sensitive DB server, the MinimumSubtotal parameter in the StoreAndLoadFromParameterizedRawSql test would fail (as Parameter turns it to lowercase).  The rest of the changes are resharper warnings.

### DIFF
--- a/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
+++ b/source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
@@ -15,19 +15,15 @@ namespace Nevermore.IntegrationTests.RelationalTransaction
         [Test]
         public void LoadWithSingleId()
         {
-            using (var trn = Store.BeginTransaction())
-            {
-                trn.Load<Product>("A");
-            }
+            using var trn = Store.BeginTransaction();
+            var _ = trn.Load<Product>("A");
         }
 
         [Test]
         public void LoadWithMultipleIds()
         {
-            using (var trn = Store.BeginTransaction())
-            {
-                trn.LoadMany<Product>(new[] { "A", "B" });
-            }
+            using var trn = Store.BeginTransaction();
+            var _ = trn.LoadMany<Product>(new[] { "A", "B" });
         }
 
         [Test]
@@ -64,19 +60,15 @@ namespace Nevermore.IntegrationTests.RelationalTransaction
         [Test]
         public void LoadWithMoreThan2100Ids()
         {
-            using (var trn = Store.BeginTransaction())
-            {
-                trn.LoadMany<Product>(Enumerable.Range(1, 3000).Select(n => "ID-" + n));
-            }
+            using var trn = Store.BeginTransaction();
+            var _ = trn.LoadMany<Product>(Enumerable.Range(1, 3000).Select(n => "ID-" + n));
         }
 
         [Test]
         public void LoadStreamWithMoreThan2100Ids()
         {
-            using (var trn = Store.BeginTransaction())
-            {
-                trn.LoadStream<Product>(Enumerable.Range(1, 3000).Select(n => "ID-" + n));
-            }
+            using var trn = Store.BeginTransaction();
+            var _ = trn.LoadStream<Product>(Enumerable.Range(1, 3000).Select(n => "ID-" + n));
         }
 
         [Test]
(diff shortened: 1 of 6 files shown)


source/Nevermore.IntegrationTests/RelationalTransaction/LoadFixture.cs
```diff
@@ -15,19 +15,15 @@ public class LoadFixture : FixtureWithRelationalStore
         [Test]
         public void LoadWithSingleId()
         {
-            using (var trn = Store.BeginTransaction())
-            {
-                trn.Load<Product>("A");
-            }
+            using var trn = Store.BeginTransaction();
```
andrew-at-octopus, line 18: This is all stylecop except for changing MinimumSubtotal to minimum_subtotal

source/Nevermore/Transient/DbCommandExtensions.cs
```diff
@@ -128,39 +84,96 @@ public static async Task<object> ExecuteScalarWithRetryAsync(this DbCommand comm
...
-                }
+                await using var validCommand = await ValidCommandAsync.For(command, connectionRetryPolicy, cancellationToken);
+                return await validCommand.ExecuteScalarAsync(cancellationToken);
             });
         }
 
         static void GuardConnectionIsNotNull(DbCommand command)
         {
-            if (command.Connection == null)
+            if (command?.Connection == null)
```
andrew-at-octopus, line 94: This is an interesting one... when this code was used as part of the EnsureValidConnection call, it verified that command was not null, then skipped it if it was... but ALL the methods call this Guard Method first... and if command was null it would have thrown a null exception error anyway.... so there is a slight difference here, in that a null command will now throw a connection property exception... should I revert this?

source/Nevermore/CommandExecutor.cs
```diff
@@ -182,17 +180,15 @@ public async Task<DbDataReader> ExecuteReaderAsync(CancellationToken cancellatio
             }
         }
 
-        public async Task<T[]> ReadResultsAsync<T>(Func<DbDataReader, T> mapper)
+        public async Task<T[]> ReadResultsAsync<T>(Func<DbDataReader, T> mapper, CancellationToken cancellationToken = default)
```
droyad, line 183: This is a breaking change to the public contract, either Major version bump or create an overload

source/Nevermore/Transient/DbConnectionExtensions.cs
```diff
@@ -18,14 +19,14 @@ public static void OpenWithRetry(this DbConnection connection, RetryPolicy retry
...
-            return OpenWithRetryAsync(connection, RetryManager.Instance.GetDefaultSqlConnectionRetryPolicy());
+            return OpenWithRetryAsync(connection, RetryManager.Instance.GetDefaultSqlConnectionRetryPolicy(), cancellationToken);
         }
 
-        public static Task OpenWithRetryAsync(this DbConnection connection, RetryPolicy retryPolicy)
+        public static Task OpenWithRetryAsync(this DbConnection connection, RetryPolicy retryPolicy, CancellationToken cancellationToken = default)
         {
-            return (retryPolicy ?? RetryPolicy.NoRetry).LoggingRetries("Open Database Connection").ExecuteActionAsync(connection.OpenAsync);
+            return (retryPolicy ?? RetryPolicy.NoRetry).LoggingRetries("Open Database Connection").ExecuteActionAsync(() => connection.OpenAsync(cancellationToken));
         }
```
droyad, line 30: These are breaking changes (see above)

droyad (approved):
Please drop this into #topic-nevermore as well

source/Nevermore/Transient/DbCommandExtensions.cs
```diff
@@ -128,39 +84,96 @@ public static async Task<object> ExecuteScalarWithRetryAsync(this DbCommand comm
...
-            GuardConnectionIsNotNull(command);
+            readonly DbCommand innerCommand = null;
+
+            ValidCommandAsync(DbCommand command, DbConnection connection = null)
+            {
+                innerCommand = command;
+                ourConnection = connection;
+            }
+            
+            public static async Task<ValidCommandAsync> For(DbCommand command, RetryPolicy retryPolicy, CancellationToken cancellationToken)
```
APErebus, line 150: nit: `ForAsync`

source/Nevermore/Transient/DbCommandExtensions.cs
```diff
@@ -83,18 +57,8 @@ public static async Task<DbDataReader> ExecuteReaderWithRetryAsync(this DbComman
...
-                    var weOwnTheConnectionLifetime = EnsureValidConnection(command, connectionRetryPolicy);
-                    try
-                    {
-                        return await command.ExecuteReaderAsync(commandBehavior, cancellationToken);
-                    }
-                    catch (Exception)
-                    {
-                        if (weOwnTheConnectionLifetime && command.Connection != null &&
-                            command.Connection.State == ConnectionState.Open)
-                            await command.Connection.CloseAsync();
```
andrew-at-octopus, line 95: Shouldn't this close code have been in a finally block, not a catch block???  Could this be keeping our connections open when this method was used, we opened the connection and there wasn't an error?

APErebus (approved):
Yeah nice! The other couple of Async calls missing CancellationTokens are I've noticed we are missing is 

`IRelationalStore.BeginReadTransactionAsync()`
`IRelationalStore.BeginWriteTransactionAsync()`

It we are doing breaking changes, would be nice to update those methods as well

source/Nevermore/Transient/DbCommandExtensions.cs
APErebus: yeah, that looks a bit erroneous :|

droyad: If you close the connection the reader won't work anymore will it?

andrew-at-octopus: Probably not... I'll check, but then I'm not sure when the connection would be closed either... following it back to the Stream method (which doesn't return the DbReader object, but a stream of the objects) it doesn't close it either... so I'm hoping that these methods don't own the connection (ever)... perhaps we should ensure that is always the case for these methods?

uglybugger:
Thank you for putting this one together :)

source/Nevermore/Transient/DbCommandExtensions.cs
andrew-at-octopus: I have reversed these changes, as they were tangengential to the main focus of the task - the use of the cancellationToken, and returned to the old pattern, as I really don't want to break the DbDataReader methods.

source/Nevermore/Transient/DbCommandExtensions.cs
andrew-at-octopus: Reverted this, mainly for the DbDataReader methods which only close in a catch block... I think there is a better pattern for this, but this PR is not the place for it.

APErebus (approved):
Looks good!

source/Nevermore/CommandExecutor.cs
APErebus: @droyad I don't think it's possible for these to be an overload because it's an optional param?

RIder indicates that the overload with the optional cancellation token would be hidden

source/Nevermore/Transient/DbConnectionExtensions.cs
andrew-at-octopus: Replaced with overloads

source/Nevermore/CommandExecutor.cs
andrew-at-octopus: CommandExecutor is an internal class, so ignoring this here

PR #148, opened 2021-06-13
Add support for `DISTINCT` select statement

zentron:
This change brings support to perform the `DISTINCT` sql operation to Nevermore. 
The approach here follows similar structure to the existing `TAKE` operation which has similar semantics albeit without the implicit ordering.


droyad (approved):
Nice! Please drop this into #topic-nevermore as a heads up

PR #149, opened 2021-06-14
Added the concept of Primary Key Handlers

slewis74:
This PR is based on the learnings from #146 and the updates being proposed in #145 for Identity primary key support.

It adds the concept of PrimaryKeyHandlers, which in many ways are similar to having TypeHandlers, but specific to the use cases required of primary keys, e.g. INSERTs and DELETEs operations.

Reviewing is probably easiest by commit. The first 2 build out the functionality and the 3rd updates the Customer class in the integration tests to use a CustomerId primary key field.


source/Nevermore.IntegrationTests/Model/StringTinyTypeIdTypeHandler.cs
```diff
@@ -0,0 +1,37 @@
+﻿#nullable enable
+using System;
+using System.Data;
+using System.Data.Common;
+using Nevermore.Advanced.TypeHandlers;
+
+namespace Nevermore.IntegrationTests.Model
+{
+    class StringTinyTypeIdTypeHandler<T> : ITypeHandler where T : TinyType<string>
```
pawelpabich, line 9: I would avoid referencing `TinyType` name anywhere in Nevermore because this might cause quite a bit of confusion. What about `CustomIdType` instead?

source/Nevermore.IntegrationTests/Model/TinyType.cs
```diff
@@ -0,0 +1,54 @@
+﻿#nullable enable
+using System;
+using System.Globalization;
+using System.Reflection;
+
+namespace Nevermore.IntegrationTests.Model
+{
+    public class TinyType<T>
```
pawelpabich, line 8: Same as above. e.g.  `CustomIdType`

source/Nevermore/IReadQueryExecutor.cs
```diff
@@ -45,6 +45,15 @@ public interface IReadQueryExecutor
...
         [Pure] TDocument Load<TDocument>(Guid id) where TDocument : class;
 
+        /// <summary>
+        /// Loads a single document given its ID. If the item is not found, returns <c>null</c>.
+        /// </summary>
+        /// <typeparam name="TDocument">The type of document being queried. Results from the database will be mapped to this type.</typeparam>
+        /// <typeparam name="TKey">The type of the Id</typeparam>
+        /// <param name="id">The <c>Id</c> of the document to find.</param>
+        /// <returns>The document, or <c>null</c> if the document is not found.</returns>
+        [Pure] TDocument Load<TDocument, TKey>(TKey id) where TDocument : class;
```
pawelpabich, line 55: We need generic members for most of the queries, I can imagine they will be added once we have a solution to the main problem?

source/Nevermore/Mapping/DocumentMap.cs
```diff
@@ -34,22 +35,12 @@ protected string TableName
         }
 
         /// <summary>
-        /// Gets or sets the prefix to be used before IDs. If you override <see cref="IdFormat"/>, this property won't
-        /// be used.
+        /// Gets or sets the primary key handler.
         /// </summary>
-        protected string IdPrefix
+        protected IPrimaryKeyHandler? PrimaryKeyHandler
```
pawelpabich, line 40: 👍

source/Nevermore/Mapping/GuidPrimaryKeyHandler.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Nevermore.Mapping
+{
+    class GuidPrimaryKeyHandler : PrimitivePrimaryKeyHandler<Guid>
+    {
+        public override object FormatKey(string tableName, int key)
```
pawelpabich, line 7: This feels wrong and only works because we actually don't call this method. If we did this cast would fail https://github.com/OctopusDeploy/Nevermore/blob/e8f71a42696af4d7cd26a65732534f900bd96cd3/source/Nevermore/Advanced/WriteTransaction.cs#L232

source/Nevermore/Mapping/IdColumnMapping.cs
```diff
@@ -19,26 +20,92 @@ internal IdColumnMapping(string columnName, Type type, IPropertyHandler handler,
...
             return this;
         }
 
+        /// <summary>
+        /// Set a function that when given the TableName will return key prefix string.
+        /// </summary>
+        /// <param name="idPrefix">The function to call back to get the prefix.</param>
+        public IIdColumnMappingBuilder Prefix(Func<string, string> idPrefix)
+        {
+            if (!(PrimaryKeyHandler is null) && Direction == ColumnDirection.FromDatabase && PrimaryKeyHandler is IIdentityPrimaryKeyHandler)
```
pawelpabich, line 70: There is a bit of type checking in a few places, I can imagine you tried to remove it but that would break too much of the existing code?

source/Nevermore/Mapping/LongPrimaryKeyHandler.cs
```diff
@@ -0,0 +1,10 @@
+namespace Nevermore.Mapping
+{
+    class LongPrimaryKeyHandler : PrimitivePrimaryKeyHandler<long>
+    {
+        public override object FormatKey(string tableName, int key)
+        {
+            return key;
```
pawelpabich, line 7: Same as Guid

source/Nevermore/Mapping/PrimitivePrimaryKeyHandler.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Nevermore.Mapping
+{
+    public abstract class PrimitivePrimaryKeyHandler<T> : IPrimitivePrimaryKeyHandler
+    {
+        public Type Type => typeof(T);
+
+        [return: NotNullIfNotNull("id")]
+        public virtual object? GetPrimitiveValue(object? id)
```
pawelpabich, line 12: `GetPrimitiveValue` convert to `ConvertToPrimitiveValue` ?

pawelpabich:
I think we are going in the direction but some abstractions don't feel fully formed.  I will try to experiment with a few approaches tomorrow.

source/Nevermore.IntegrationTests/Model/StringTinyTypeIdTypeHandler.cs
slewis74: yep 👍

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -540,20 +577,23 @@ protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand prepar
         {
             var mapping = configuration.DocumentMaps.Resolve(typeof(TDocument));
 
+            if (mapping.IdColumn is null)
```
pawelpabich, line 580: We could mark IdColumn as not nullable  if owe passed it to DocumentMap via the constructor. I can imagine this will break gazillion things in the server? https://github.com/OctopusDeploy/Nevermore/blob/e21a19848c32e44a0038fbd80b3190c3d495d14f/source/Nevermore/Mapping/DocumentMap.cs#L258

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -135,24 +135,24 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
         }
 
         public void Delete<TDocument>(string id, DeleteOptions options = null) where TDocument : class
-            => Delete<TDocument>((object) id, options);
+            => Delete<TDocument, string>(id, options);
 
         public void Delete<TDocument>(int id, DeleteOptions options = null) where TDocument : class
-            => Delete<TDocument>((object) id, options);
+            => Delete<TDocument, int>(id, options);
```
pawelpabich, line 141: Nice :)

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -208,36 +209,51 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
...
 
-        string AllocateId(string tableName, Func<int, string> idFormatter)
+        object AllocateIdUsingHandler(DocumentMap mapping)
+        {
+            if (mapping.IdColumn is null || mapping.IsIdentityId)
+                throw new InvalidOperationException($"Cannot allocate an id when an Id column has not been mapped.");
+            return mapping.IdColumn.PrimaryKeyHandler.GetNextKey(keyAllocator, mapping.TableName);
+        }
+
+        public string AllocateId(string tableName, string idPrefix)
```
pawelpabich, line 247: I was wondering if we could get rid of this method and then found this 
image

pawelpabich: What do you think about marking this method as obsolete?

source/Nevermore/Mapping/IPrimaryKeyHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        object? ConvertToPrimitiveValue(object? id);
+
+        /// <summary>
+        /// Get the next key for the given table.
+        /// </summary>
+        /// <remarks>The keyAllocator has to be passed here as it is tied to the RelationalStore instance, and thus can't be specified at configuration time in the constructor.</remarks>
+        /// <param name="keyAllocator">The key allocator to use getting the next id from.</param>
+        /// <param name="tableName">The table name the key is required for.</param>
+        /// <returns>The next key, as the type that matches the model object's Id property type. ConvertToPrimitiveValue should be called with this value if it is to be used as a Sql parameter.</returns>
+        object GetNextKey(IKeyAllocator keyAllocator, string tableName);
```
pawelpabich, line 26: This abstraction is a bit leaky. Ideally, `keyAllocator` and `tableName` would be treated as implementation details of a given primary key handler. That being said, this is not a 5 minute job so happy to keep it as a future ToDo item :)

pawelpabich (approved):
You had to jump over a lot hurdles but we are in much better place now. There is still work to do but that's ok. Thanks!

slewis74:
[This page](https://github.com/OctopusDeploy/Nevermore/wiki/Documents) should be updated when this PR merges. A new page for `Primary Key Handlers` should also be added to the Wiki.

PR #151, opened 2021-06-18
Adding GroupBy Support 

andrew-at-octopus:
Nevermore is based on a SQL database, but is missing some operations that can make queries simpler to write, and faster to run.
One missing feature is the GroupBy clause for obtaining aggregate information from the datastore.
While I was at it, I added a CountColumn helper method, which allows the (GroupBy) query to add a count column - mainly because the current Count() method expects that it finalises the query and is the only column returned (runs as scalar), whereas we want to include the count as one of the columns returned.

NOTE: Using GroupBy can easily result in queries that fall over, and at the moment there is no safeguard as to what columns are selected (eg, included in the group, or an aggregate), so be careful using it.

### DIFF
--- a/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
+++ b/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
@@ -194,6 +194,7 @@ ORDER BY [CorrelationId]";
                 .Where("IsActive = 1")
                 .OrderBy("Created");
 
+            // Note, this is less than ideal, Hint should be multi-add, and perhaps with an enum??
             var accounts = CreateQueryBuilder<object>("Accounts").Hint("WITH (UPDLOCK)");
 
             var actual = orders.InnerJoin(customers.Subquery())
@@ -258,7 +259,7 @@ WHERE ([Price] > 5)";
         }
 
         [Test]
-        public void ShouldGenerateCountBySpaceForQueryBuilder()
+        public void ShouldGenerateGroupBySpaceForQueryBuilder()
         {
             string actual = null;
             transaction.Stream<object>(Arg.Do<string>(s => actual = s), Arg.Any<CommandParameterValues>());
@@ -267,7 +268,7 @@ WHERE ([Price] > 5)";
                 .NoLock()
                 .GroupBy("SpaceId")
                 .Column("SpaceId")
-                .CountColumn("OrderCount")
+                .CalculatedColumn("COUNT (*)", "OrderCount")
                 .ToList();
 
             const string expected = @"SELECT [SpaceId],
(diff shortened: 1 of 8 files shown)


source/Nevermore/Querying/AST/Select.cs
```diff
@@ -23,8 +25,12 @@ public string GenerateSql()
         {
             var orderByString = orderBy != null ? $@"
 {orderBy.GenerateSql()}" : string.Empty;
+            
+            var groupByString = groupBy != null ? $@"
+{groupBy.GenerateSql()}" : string.Empty;
```
emilol, line 30: since `null` interpolates to empty string, could we simplify this to `var groupByString = groupBy?.GenerateSql()`

image

source/Nevermore/Querying/AST/Select.cs
```diff
@@ -7,14 +7,16 @@ public class Select : ISelect
         readonly ISelectSource from;
         readonly Where where;
         readonly OrderBy orderBy; // Can be null
+        readonly GroupBy groupBy; // Can be null
 
-        public Select(IRowSelection rowSelection, ISelectColumns columns, ISelectSource from, Where where, OrderBy orderBy)
+        public Select(IRowSelection rowSelection, ISelectColumns columns, ISelectSource from, Where where, GroupBy groupBy, OrderBy orderBy)
```
emilol, line 12: This class and its friends are public - does that mean they're part of our public API for nevermore? 

If so, as is this would be a breaking change, so we'll either need to add backwards compat, or do a major version bump.

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
```diff
@@ -257,6 +257,48 @@ public void ShouldGenerateCountForQueryBuilder()
             actual.Should().Be(expected);
         }
 
+        [Test]
+        public void ShouldGenerateCountBySpaceForQueryBuilder()
```
emilol, line 261: Good to see test coverage, makes it super clear what we're adding 👍

source/Nevermore/Advanced/QueryBuilder.cs
```diff
@@ -113,6 +113,20 @@ public IQueryBuilder<TRecord> CalculatedColumn(string expression, string columnA
             return this;
         }
 
+
+        public IQueryBuilder<TRecord> CountColumn(string columnAlias)
+        {
+            selectBuilder.AddColumnSelection(new AliasedColumn(new CalculatedColumn(new CustomExpression("COUNT (*)")), columnAlias));
```
emilol, line 119: we could reuse the overload that we're decorating here, to minimise our reasons to change.
```suggestion
            selectBuilder.CalculatedColumn("COUNT (*)", columnAlias);
```

source/Nevermore/Advanced/QueryBuilder.cs
```diff
@@ -113,6 +113,20 @@ public IQueryBuilder<TRecord> CalculatedColumn(string expression, string columnA
...
+            return this;
+        }
+
+        public IQueryBuilder<TRecord> CountColumn(string expression, bool distinct, string columnAlias)
+        {
+            var distinctClause = distinct ? "DISTINCT " : string.Empty;
+            selectBuilder.AddColumnSelection(new AliasedColumn(new CalculatedColumn(new CustomExpression($"COUNT ({distinctClause}{expression})")), columnAlias));
```
emilol, line 126: as above
```suggestion
            CalculatedColumn($"COUNT ({distinctClause}{expression})", columnAlias);
```

emilol (approved):
LGTM - one key question about versioning/compat to sort out before merging, and a few minor nitpicks that shouldn't block this PR.

source/Nevermore/Querying/AST/Select.cs
andrew-at-octopus: If I move the carriage return to the GenerateSql method yes.  Done

PR #152, opened 2021-06-21
Removed string based API from the mapping API

pawelpabich:
This PR builds on top of https://github.com/OctopusDeploy/Nevermore/pull/149 and changes 2 things:

1. It fixes a problem where a wrong `KeyHandler` was assigned for documents with custom prefix
2. Now that Ids can be of a different type than string we should remove string based APIs like `Format` and `Prefix` and move where they belong, which is specific `KeyHandler`s 

This is a breaking change. That being  said Octofront doesn't use `IdPrefix` and most of the cases in the server can be simply removed as the duplicate what Nevermore already does internally.

image


source/Nevermore/Mapping/PrimitivePrimaryKeyHandler.cs
```diff
@@ -6,7 +6,7 @@ namespace Nevermore.Mapping
 {
     public abstract class PrimaryKeyHandler<T> : IPrimaryKeyHandler
     {
-        public Type Type => typeof(T);
+        public virtual Type Type => typeof(T);
```
slewis74, line 9: Does this need to be virtual? It is only meant to represent the underlying type, so I'm not sure it'd be a good idea for an inheriting class to override it.

slewis74 (approved):
Very nice, I was trying to stay too close to the original idea for IdPrefix etc, this is much tidier 👍 

One question inline about a `virtual` on the Type property.

PR #153, opened 2021-06-23
Support more than 1000 related documents

laurencee:
## Summary

In Hosted we can have at least as many related documents as there are hosted instances which means nevermore needs to be able to deal with linking thousands of related documents.

Required for this performance change: https://github.com/OctopusDeploy/Octofront/pull/2858

## Restrictions dealt with in this PR:
~~* Insert statement batches [allow up to 1000 entries at a time](https://docs.microsoft.com/en-us/sql/t-sql/queries/table-value-constructor-transact-sql?view=sql-server-ver15#limitations-and-restrictions), so we have to split statements.~~
~~* SQL Commands [allow up to 2100 parameters](https://docs.microsoft.com/en-us/sql/sql-server/maximum-capacity-specifications-for-sql-server?view=sql-server-ver15#-objects) each (-2 as ADO.NET commands run through `sp_execsql` which takes 2 parameters) so we must also split our calls into multiple commands as necessary.~~
- @flin-8 I've just solved the issue with related documents (caused by SQL Server's 2100 parameter limit) by using a [table value parameter](https://github.com/OctopusDeploy/Nevermore/wiki/Table-Valued-Parameters#bulk-inserts). I've reverted the changes with multiple commands since I don't think it's a high value change (plus it involves more changes).

The bulk of changes are in `DataModificationQueryBuilder.cs `

## Performance

Table valued parameter is faster once we are above 100 related documents for INSERT's. It's pretty much always faster for UPDATE's.

**Explicit parameters**

|                                 Method |        Mean |     Error |    StdDev |      Median |         P50 |         P80 |         P90 |         P95 |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------------------- |------------:|----------:|----------:|------------:|------------:|------------:|------------:|------------:|-----------:|----------:|------:|----------:|
|  Insert100OrdersWith10RelatedDocuments |    44.44 ms |  0.798 ms |  1.576 ms |    44.01 ms |    44.01 ms |    45.71 ms |    46.55 ms |    47.90 ms |   636.3636 |  181.8182 |     - |   5.66 MB |
|  Insert100OrdersWith20RelatedDocuments |    60.63 ms |  1.451 ms |  3.973 ms |    59.77 ms |    59.77 ms |    63.26 ms |    65.48 ms |    67.35 ms |  1000.0000 |  333.3333 |     - |   8.17 MB |
|  Insert100OrdersWith50RelatedDocuments |    91.63 ms |  1.739 ms |  4.065 ms |    89.88 ms |    89.88 ms |    93.13 ms |    98.34 ms |    99.84 ms |  1800.0000 |  800.0000 |     - |  15.11 MB |
|  Insert100OrdersWith80RelatedDocuments |   137.28 ms |  2.666 ms |  4.380 ms |   135.45 ms |   135.45 ms |   140.69 ms |   144.79 ms |   146.19 ms |  2750.0000 |  500.0000 |     - |  22.49 MB |
| **Insert100OrdersWith100RelatedDocuments** |   171.01 ms |  3.335 ms |  4.782 ms |   168.62 ms |   168.62 ms |   174.14 ms |   180.53 ms |   180.61 ms |  3333.3333 |  666.6667 |     - |  26.78 MB |
| Insert100OrdersWith500RelatedDocuments | 1,398.26 ms | 24.616 ms | 26.338 ms | 1,390.64 ms | 1,390.64 ms | 1,422.44 ms | 1,429.40 ms | 1,438.93 ms | 12000.0000 | 2000.0000 |     - | 102.56 MB |
|  Update100OrdersWith10RelatedDocuments |   628.17 ms | 12.487 ms | 32.895 ms |   620.65 ms |   620.65 ms |   658.09 ms |   675.60 ms |   681.59 ms |          - |         - |     - |   6.48 MB |
|  Update100OrdersWith20RelatedDocuments |   438.76 ms |  8.861 ms | 25.708 ms |   426.47 ms |   426.47 ms |   465.17 ms |   480.11 ms |   493.37 ms |  1000.0000 |         - |     - |    8.9 MB |
|  Update100OrdersWith50RelatedDocuments |   564.09 ms |  5.723 ms |  5.073 ms |   563.82 ms |   563.82 ms |   568.41 ms |   570.11 ms |   571.66 ms |  1000.0000 |         - |     - |  15.58 MB |
|  Update100OrdersWith80RelatedDocuments |   871.80 ms |  8.539 ms |  7.570 ms |   869.53 ms |   869.53 ms |   878.53 ms |   882.99 ms |   884.30 ms |  2000.0000 |         - |     - |   22.8 MB |
| Update100OrdersWith100RelatedDocuments |   895.45 ms |  3.637 ms |  3.402 ms |   895.59 ms |   895.59 ms |   898.16 ms |   899.84 ms |   900.48 ms |  3000.0000 |         - |     - |  27.08 MB |
| Update100OrdersWith300RelatedDocuments | 2,629.34 ms | 39.194 ms | 30.600 ms | 2,617.59 ms | 2,617.59 ms | 2,645.72 ms | 2,648.84 ms | 2,677.52 ms |  8000.0000 | 1000.0000 |     - |  70.28 MB |
| Update100OrdersWith500RelatedDocuments | 3,375.24 ms | 28.178 ms | 22.000 ms | 3,377.39 ms | 3,377.39 ms | 3,385.88 ms | 3,396.41 ms | 3,406.83 ms | 12000.0000 | 2000.0000 |     - | 101.21 MB |
| Update100OrdersWith700RelatedDocuments | 5,232.84 ms | 68.633 ms | 60.842 ms | 5,218.74 ms | 5,218.74 ms | 5,283.34 ms | 5,299.94 ms | 5,329.17 ms | 17000.0000 | 2000.0000 |     - | 137.74 MB |

**Table valued parameter**

|                                 Method |        Mean |     Error |    StdDev |      Median |         P50 |         P80 |         P90 |         P95 |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------------------- |------------:|----------:|----------:|------------:|------------:|------------:|------------:|------------:|-----------:|----------:|------:|----------:|
|  Insert100OrdersWith10RelatedDocuments |    84.91 ms |  1.683 ms |  2.518 ms |    83.95 ms |    83.95 ms |    87.10 ms |    88.74 ms |    89.86 ms |   666.6667 |         - |     - |   5.68 MB |
|  Insert100OrdersWith20RelatedDocuments |    93.65 ms |  0.818 ms |  0.639 ms |    93.54 ms |    93.54 ms |    94.01 ms |    94.17 ms |    94.65 ms |   833.3333 |  166.6667 |     - |   7.53 MB |
|  Insert100OrdersWith50RelatedDocuments |   115.63 ms |  2.310 ms |  2.161 ms |   115.38 ms |   115.38 ms |   117.64 ms |   118.16 ms |   118.88 ms |  1600.0000 |  200.0000 |     - |  12.88 MB |
|  Insert100OrdersWith80RelatedDocuments |   140.24 ms |  2.767 ms |  3.598 ms |   139.42 ms |   139.42 ms |   142.89 ms |   145.88 ms |   146.94 ms |  2250.0000 |  500.0000 |     - |  18.45 MB |
| **Insert100OrdersWith100RelatedDocuments** |   160.90 ms |  2.269 ms |  2.870 ms |   159.88 ms |   159.88 ms |   161.68 ms |   162.77 ms |   166.60 ms |  2666.6667 |  666.6667 |     - |  21.69 MB |
| Insert100OrdersWith500RelatedDocuments |   413.59 ms |  8.231 ms | 17.893 ms |   406.68 ms |   406.68 ms |   437.92 ms |   442.73 ms |   444.15 ms | 10000.0000 | 3000.0000 |     - |  82.33 MB |
|  Update100OrdersWith10RelatedDocuments |   317.83 ms |  3.105 ms |  2.752 ms |   317.48 ms |   317.48 ms |   320.54 ms |   321.55 ms |   321.90 ms |          - |         - |     - |   6.42 MB |
|  Update100OrdersWith20RelatedDocuments |   360.25 ms |  3.591 ms |  3.184 ms |   359.56 ms |   359.56 ms |   363.01 ms |   364.13 ms |   365.34 ms |  1000.0000 |         - |     - |   8.23 MB |
|  Update100OrdersWith50RelatedDocuments |   513.87 ms |  5.266 ms |  4.669 ms |   513.88 ms |   513.88 ms |   517.25 ms |   520.22 ms |   520.94 ms |  1000.0000 |         - |     - |   13.4 MB |
|  Update100OrdersWith80RelatedDocuments |   768.64 ms |  8.514 ms |  7.964 ms |   767.06 ms |   767.06 ms |   775.43 ms |   778.55 ms |   780.93 ms |  2000.0000 |         - |     - |  18.96 MB |
| Update100OrdersWith100RelatedDocuments |   843.89 ms | 12.186 ms | 11.399 ms |   846.96 ms |   846.96 ms |   853.57 ms |   857.09 ms |   858.15 ms |  2000.0000 |         - |     - |  22.19 MB |
| Update100OrdersWith300RelatedDocuments | 2,167.08 ms | 21.294 ms | 19.919 ms | 2,174.41 ms | 2,174.41 ms | 2,178.31 ms | 2,186.26 ms | 2,195.24 ms |  7000.0000 | 1000.0000 |     - |   56.1 MB |
| Update100OrdersWith500RelatedDocuments | 3,684.23 ms | 17.933 ms | 16.775 ms | 3,687.27 ms | 3,687.27 ms | 3,694.90 ms | 3,697.50 ms | 3,704.52 ms | 10000.0000 | 3000.0000 |     - |  82.67 MB |
| Update100OrdersWith700RelatedDocuments | 5,206.59 ms | 38.416 ms | 34.054 ms | 5,198.03 ms | 5,198.03 ms | 5,229.90 ms | 5,247.67 ms | 5,262.59 ms | 13000.0000 | 2000.0000 |     - | 111.21 MB |


source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +331,158 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
+                return commands.ToArray();
+            }
 
             var sb = new StringBuilder();
-            sb.AppendLine(statement);
+            sb.AppendLine(updateStatement);
             sb.AppendLine();
 
+            // sql commands are executed inside sp_executesql which means table variables and local temp tables won't live between commands
+            var globalTempTableName = $"##references_{Guid.NewGuid().ToString().Replace('-', '_')}";
```
laurencee, line 434: This is the best solution I could come up with to allow this approach to work over multiple commands.

Global temp tables are effectively reference counted and only exist while there are sessions open with references to the global temp table, so cleanup is automatically taken care of.

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +331,153 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
+                return commands.ToArray();
+            }
 
             var sb = new StringBuilder();
-            sb.AppendLine(statement);
+            sb.AppendLine(updateStatement);
             sb.AppendLine();
 
+            // sql commands are executed inside sp_executesql which means table variables and local temp tables won't live between commands
+            var globalTempTableName = $"##{configuration.RelatedDocumentsGlobalTempTableNameGenerator()}";
```
laurencee, line 429: This is the best solution I could come up with to allow this update approach to work over multiple commands.

Global temp tables are effectively reference counted and only exist while there are sessions open with references to the global temp table, so cleanup is automatically taken care of.

laurencee: An option to be consistent with the previous version is to generate a different SQL statement here if we only need to create a single command.
If we only need a single command we can use a table variable, otherwise we need the global temp table.

laurencee:
The large number of new lines is from an Assent file. The code changes are around +250, -130

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -50,13 +54,17 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
 
         public async Task InsertAsync<TDocument>(TDocument document, InsertOptions options, CancellationToken cancellationToken = default) where TDocument : class
         {
-            var command = builder.PrepareInsert(new[] {document}, options);
-            await configuration.Hooks.BeforeInsertAsync(document, command.Mapping, this);
+            var commands = builder.PrepareInsert(new[] {document}, options);
```
flin-8, line 57: I actually don't know the answer to this: should we loop through _all_ commands  for the `BeforeInsertAsync` and `AfterInsertAsync` hooks? rather than processing each command one by one?

Actually - maybe we should only run the hook once. The mapping for the commands should all be the same right?

laurencee: Great question! I don't know the answer to this.

source/Nevermore.IntegrationTests/RelatedDocumentTableFixture.cs
```diff
@@ -171,10 +170,20 @@ public void Read()
                 .BDDfy();
         }
 
-        [Test]
-        public void Insert()
+        [TestCase(1)]
+        [TestCase(1000)]
+        [TestCase(2001)]
+        [TestCase(2096)] // exact SQL Server param per command limit
```
flin-8, line 176: Probably makes sense to add a case for `2097` then? 1 above the limit

source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.UpdateWithManyRelatedDocuments.approved.txt
```diff
@@ -1,24 +1,18 @@
 UPDATE [dbo].[TestDocumentTbl]  SET [AColumn] = @AColumn, [JSON] = @JSON WHERE [Id] = @Id
 
 DECLARE @references as TABLE (Reference nvarchar(400), ReferenceTable nvarchar(400))
-
-DELETE FROM @references
```
flin-8, line 5: Why don't we need this `DELETE` statement anymore? was it doing nothing before?

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -34,20 +38,21 @@ public DataModificationQueryBuilder(IRelationalStoreConfiguration configuration,
             this.keyAllocator = keyAllocator;
         }
 
-        public PreparedCommand PrepareInsert(IReadOnlyList<object> documents, InsertOptions options = null)
+        public PreparedCommand[] PrepareInsert(IReadOnlyList<object> documents, InsertOptions options = null)
         {
+            var commands = new List<PreparedCommand>();
```
flin-8, line 43: Do we need this list here? we're only adding to it once

source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.UpdateWithManyRelatedDocuments.approved.txt
laurencee: Yep, did nothing as there weren't any records in the table variable at this point.

source/Nevermore/Util/DataModificationQueryBuilder.cs
laurencee: Nice pickup, we don't need it at all actually as all the commands are now created in 1 place.

source/Nevermore/Advanced/WriteTransaction.cs
pawelpabich: > Actually - maybe we should only run the hook once. 

This makes sense to me. The fact that we might have one or more commands is an implementation detail. As a hook writer I would expect to get called once per document.

> The mapping for the commands should all be the same right?
We could make this assumption or decouple the code from the builder and retrieve the map directly from `configuration.DocumentMaps.Resolve()`.

source/Nevermore/RelationalStoreConfiguration.cs
```diff
@@ -50,6 +50,8 @@ public RelationalStoreConfiguration(Func<string> connectionStringFunc)
 
             QueryLogger = new DefaultQueryLogger();
 
+            RelatedDocumentsGlobalTempTableNameGenerator = () => Guid.NewGuid().ToString().Replace('-', '_');
```
pawelpabich, line 53: You could use Guid.ToString("N") to remove separator characters.

laurencee: TIL `Guid.ToString()` has its own options ;)

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +329,160 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
-                    var relatedTableName = related[x].relatedTableName;
-
-                    var relatedVariableName = relatedVariablePrefix + x;
-                    parameters.Add(relatedVariableName, relatedDocumentId);
-                    if (x > 0)
-                        sb.Append(",");
-                    sb.AppendLine($"(@{parentIdVariable}, '{mapping.TableName}', @{relatedVariableName}, '{relatedTableName}')");
+                    var reachedParameterLimit = false;
+
+                    while (!reachedParameterLimit && remaining > 0)
```
flin-8, line 360: It looks like we are adding 1 new parameter per related document? So could we split and process each batch separately rather than using a while loop?

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +329,160 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
             sb.AppendLine();
 
-            if (relatedDocumentData.Any(d => d.Related.Any()))
-                sb.AppendLine("DECLARE @references as TABLE (Reference nvarchar(400), ReferenceTable nvarchar(400))");
+            string tableVariableName = "@references";
+            if (parameters.Count + relatedDocumentData.Sum(x => x.Related.Length) > SqlServerParameterLimit) // we need multiple sql commands
+            {
+                // sql commands are executed inside sp_executesql which means table variables and local temp tables won't live between commands
+                tableVariableName = $"##{configuration.RelatedDocumentsGlobalTempTableNameGenerator()}";
+                sb.AppendLine($"CREATE TABLE {tableVariableName} (Reference nvarchar(400) COLLATE SQL_Latin1_General_CP1_CS_AS, ReferenceTable nvarchar(400) COLLATE SQL_Latin1_General_CP1_CS_AS)").AppendLine();
```
flin-8, line 431: Why specify collation here?

laurencee: When I run this locally without I get a collation exception if I don't specify the collation here. Looking at my `tempdb` the collation is `Latin1_General_CI_AS` while my actual databases are all `SQL_Latin1_General_CP1_CI_AS`.

Maybe the collation needs to be configurable in the configuration in case someone else using this happens to have a different collation on their databases?

source/Nevermore/Advanced/WriteTransaction.cs
flin-8: Ah it's resolved by `DataModificationQueryBuilder.GetMapping()`

source/Nevermore/Util/DataModificationQueryBuilder.cs
laurencee: I don't really follow what you're asking here. What's the improvement over the batching structure that exists here?

On a side note, this function doesn't need the second while loop defined anymore as I no longer use the `reachedParameterLimit` variable so that drops a few lines of code.

source/Nevermore/Util/DataModificationQueryBuilder.cs
flin-8: So it looks like the column collation in the temporary table needs to match the collation from the related document table column when performing the join. The reason why `SQL_Latin1_General_CP1_CS_AS` is there is because that's how we set up the test database.

We would either need to:
1. query the database schema to determine what the collation of the related document table columns are, or
2. find a way around not using global temp tables

source/Nevermore/Util/DataModificationQueryBuilder.cs
flin-8: Sorry should've been more specific: I'm thinking that readability/understandability would be improved by not using a while loop.

Could you push up those changes now anyway?

laurencee: Pushed

source/Nevermore/Util/DataModificationQueryBuilder.cs
laurencee: Pushed change with collation defined in configuration

source/Nevermore/IRelationalStoreConfiguration.cs
```diff
@@ -54,15 +55,27 @@ public interface IRelationalStoreConfiguration
...
+        /// Defines a name generator to be used when creating global temp tables during related document updates. <para />
+        /// This is used when there are more related documents than can fit in a single SQL command
+        /// </summary>
+        Func<string> RelatedDocumentsGlobalTempTableNameGenerator { get; }
+
+        /// <summary>
+        /// Used with <see cref="RelatedDocumentsGlobalTempTableNameGenerator"/> when transferring data from the temp table to the related documents table to ensure collation matches. <para />
+        /// Defaults to SQL_Latin1_General_CP1_CS_AS
+        /// </summary>
+        public string RelatedDocumentsDatabaseCollation { get; set; }
```
laurencee, line 79: See previous discussion: https://github.com/OctopusDeploy/Nevermore/pull/153#discussion_r659399751

> 
> 
> So it looks like the column collation in the temporary table needs to match the collation from the related document table column when performing the join. The reason why `SQL_Latin1_General_CP1_CS_AS` is there is because that's how we set up the test database.
> 
> We would either need to:
> 
>     1. query the database schema to determine what the collation of the related document table columns are, or
> 
>     2. find a way around not using global temp tables

source/Nevermore/Util/DataModificationQueryBuilder.cs
laurencee: Continued https://github.com/OctopusDeploy/Nevermore/pull/153/files#r659409108

source/Nevermore/Util/DataModificationQueryBuilder.cs
laurencee: Closing this conversation so it's not confusing looking at old code with whatever suggestion is made.

source/Nevermore/IRelationalStoreConfiguration.cs
flin-8: I'm not convinced static configuration is suitable for this - it's not great experience for the caller to set this.

What if we injected a func that lazily queried the collation from the database?

We do something similar with the key allocator already (make DB calls during query text generation)
https://github.com/OctopusDeploy/Nevermore/blob/516d046eb4f427c442ccd7c4b6d7e15bc597cd98/source/Nevermore/Util/DataModificationQueryBuilder.cs#L33

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +329,155 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
+            InsertOptions options)
         {
+            var commands = new List<PreparedCommand>();
+            var parameters = new CommandParameterValues();
+            // we need to create these document parameters for each command - parameters are disposed at the end of execution in 'CommandExecutor'
+            Func<CommandParameterValues> documentParametersCreator = () => GetDocumentParameters(m => keyAllocator(m), options.CustomAssignedId, documents, mapping, DataModification.Insert);
+            parameters.AddRange(documentParametersCreator());
+
             var relatedDocumentData = GetRelatedDocumentTableData(mapping, documents);
+            if (!relatedDocumentData.Any(x => x.Related.Any()))
```
pawelpabich, line 345: `!relatedDocumentData.Any(x => x.Related.Any()` => `if (!RelatedDocumentDataPresent)` ?. It's used in a few places and this reads better for me

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +329,155 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
             foreach (var data in relatedDocumentData.Where(g => g.Related.Length > 0))
             {
                 var relatedVariablePrefix = $"{data.TableName.ToLower()}_";
 
-                sb.AppendLine($"INSERT INTO [{data.SchemaName}].[{data.TableName}] ([{data.IdColumnName}], [{data.IdTableColumnName}], [{data.RelatedDocumentIdColumnName}], [{data.RelatedDocumentTableColumnName}]) VALUES");
-                var related = data.Related;
-
-                for (var x = 0; x < related.Length; x++)
+                var remaining = data.Related.Length;
+                while (remaining > 0)
```
pawelpabich, line 356: I agree with Frank here. The while loop is hard to understand. What about this high level approach:

`relatedDocumentToInsert.SplitIntoBatches((FirstBatchSize, commandWithExistingInsertSQL), (OtherBatchSizes, new Command())).Select((command, batch) => add batch to the command).ToListOfCommands()`.  I would leave this one until the very end as it is just the refactor.

source/Nevermore/IRelationalStoreConfiguration.cs
pawelpabich: What about populating `@reference` table multiple times, once per command?

pawelpabich: > query the database schema to determine what the collation of the related document table columns are, or

I don't think Nevermore make this type of queries anywhere so existing sql users might not have enough permissions.

pawelpabich:
> In Hosted we can have at least as many related documents as there are hosted instances which means nevermore needs to be able to deal with linking thousands of related documents.
 
 
 Can you link to the PR in Hosted Portal that explains why we are even doing this? E.g. perf improvements

pawelpabich:
It might be a good idea to create a branch in Octopus Server with your current changes and see if any tests fail.

source/Nevermore/IRelationalStoreConfiguration.cs
laurencee: > 
> 
> What about populating `@reference` table multiple times, once per command?

Not enough parameters in the command available to do this.

flin-8: what id we create a real table in the database?

laurencee: We can query for the collation information instead of having it as a configuration value, but this requires `VIEW SERVER STATE` or `VIEW DATABASE STATE` permissions which doesn't seem unreasonable.

laurencee:
> 
> 
> It might be a good idea to create a branch in Octopus Server with your current changes and see if any tests fail.

Yep, it all passes with this branch version of Nevermore.

source/Nevermore/IRelationalStoreConfiguration.cs
laurencee: I was able to determine the collation dynamically using the above SQL. 
It now works regardless of the database column collations as long as the executor has the permission mentioned above.

https://github.com/OctopusDeploy/Nevermore/pull/153/files#diff-94be40892362f618a2307a0d89bf656db3bcf5ab10f60619ac87954a1800647aR492

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +329,177 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
+            return commands.ToArray();
+        }
+
+        void AssertAllowedSplitCommands(int currentParameterCount)
+        {
+            if (!configuration.EnableSplittingCommands)
+                throw new InvalidOperationException("Too many parameters to fit into a single SQL Command (limit 2100): " + (currentParameterCount + 2));
+        }
+
+        string CreateGlobalTempTable(StringBuilder sb)
```
laurencee, line 492: Dynamic collation detection

source/Nevermore/Advanced/WriteTransaction.cs
laurencee: This has been updated now so hooks only get called once per document and the mapping has been pulled up to the `WriteTransaction` level.

https://github.com/OctopusDeploy/Nevermore/pull/153/commits/01bac4a56aaa403c90a91d79600fc47de6805d86

source/Nevermore/Util/DataModificationQueryBuilder.cs
flin-8: @pawelpabich there's a bug with INSERTS and UPDATES that doesn't handle multiple related documents. I've pushed up a commit to expose this: https://github.com/OctopusDeploy/Nevermore/pull/153/commits/5ad88f8a0c21bef614bb995f7d27aaa1177a6670

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -330,87 +325,177 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, string> allocateI
...
+                        {
+                            commands.Add(new PreparedCommand(sb.ToString(), parameters, RetriableOperation.Update, mapping, updateOptions.CommandTimeout));
+                            sb = new StringBuilder();
+                            parameters = parametersCreator();
+                        }
+                    }
 
                     sb.AppendLine($"DELETE FROM [{data.SchemaName}].[{data.TableName}] WHERE [{data.IdColumnName}] = @{IdVariableName}");
-                    sb.AppendLine($"    AND [{data.RelatedDocumentIdColumnName}] not in (SELECT Reference FROM @references)");
+                    sb.AppendLine($"AND [{data.RelatedDocumentIdColumnName}] not in (SELECT Reference FROM {tableVariableName})");
```
flin-8, line 465: @pawelpabich You mentioned table valued parameters before, I think we misunderstood how you would use it. This might actually solve our problem. I'm going to give it a go

https://github.com/OctopusDeploy/Nevermore/wiki/Table-Valued-Parameters#bulk-inserts

pawelpabich: 👍

pawelpabich: Happy to review the PR once you have something working.

source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.InsertSingleDocumentWithManyRelatedDocuments.approved.txt
```diff
@@ -1,11 +1,9 @@
...
+INSERT INTO [dbo].[RelatedDocument] ([Id], [Table], [RelatedDocumentId], [RelatedDocumentTable])
+SELECT [Id], [Table], [RelatedDocumentId], [RelatedDocumentTable] FROM @0__relatedDocumentTableValuedParameter
 
 @Id=New-Id
 @JSON={}
 @AColumn=AValue
-@relateddocument_0=Rel-1
-@relateddocument_1=Rel-2
\ No newline at end of file
+@0__relatedDocumentTableValuedParameter=Nevermore.TableValuedParameter
```
flin-8, line 9: I haven't bothered to remove the prefix if there's only one - do we think this is important?

source/Nevermore/UpgradeScripts/Script0003-RelatedDocumentType.sql
```diff
@@ -0,0 +1 @@
+﻿CREATE TYPE dbo.RelatedDocumentTableValuedParameter AS TABLE ([Id] NVARCHAR(400), [Table] NVARCHAR(400), [RelatedDocumentId] NVARCHAR(400), [RelatedDocumentTable] NVARCHAR(400))
```
flin-8, line 1: The type for the table valued parameter needs to be created upfront. We have prior art here with [ParameterList](https://github.com/OctopusDeploy/Nevermore/blob/d50d9c0dd50cadd468cad1dfcfddb271586b2c14/source/Nevermore/UpgradeScripts/Script0002-ParameterList.sql#L2-L1)

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -384,25 +385,14 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
...
+                parameters.AddTable(tableVariableName, CreateRelatedDocumentTableValuedParameter(mapping, data));
 
-                    var relatedVariableName = relatedVariablePrefix + x;
-                    parameters.Add(relatedVariableName, relatedDocumentId);
-                    if (x > 0)
-                        sb.Append(",");
-                    sb.AppendLine($"(@{parentIdVariable}, '{mapping.TableName}', @{relatedVariableName}, '{relatedTableName}')");
-                }
+                sb.AppendLine($"INSERT INTO [{data.SchemaName}].[{data.TableName}] ([{data.IdColumnName}], [{data.IdTableColumnName}], [{data.RelatedDocumentIdColumnName}], [{data.RelatedDocumentTableColumnName}])");
+                sb.AppendLine($"SELECT [{data.IdColumnName}], [{data.IdTableColumnName}], [{data.RelatedDocumentIdColumnName}], [{data.RelatedDocumentTableColumnName}] FROM @{tableVariableName}");
```
flin-8, line 395: Instead of supplying many variables, just SELECT from `@{tableVariableName}`

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -420,34 +410,21 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
...
-                    var valueBlocks = data.Related.Select((r, idx) => $"(@{relatedVariablePrefix}{idx}, '{r.relatedTableName}')");
-                    sb.Append("INSERT INTO @references VALUES ");
-                    sb.AppendLine(string.Join(", ", valueBlocks));
-                    sb.AppendLine();
+                    var tableVariableName = $"{index++}__relatedDocumentTableValuedParameter";
+                    parameters.AddTable(tableVariableName, CreateRelatedDocumentTableValuedParameter(mapping, data));
 
                     sb.AppendLine($"DELETE FROM [{data.SchemaName}].[{data.TableName}] WHERE [{data.IdColumnName}] = @{IdVariableName}");
-                    sb.AppendLine($"    AND [{data.RelatedDocumentIdColumnName}] not in (SELECT Reference FROM @references)");
+                    sb.AppendLine($"    AND [{data.RelatedDocumentIdColumnName}] not in (SELECT [{data.RelatedDocumentIdColumnName}] FROM @{tableVariableName})");
```
flin-8, line 422: We can use our table valued parameter right here, instead of populating a temp table 😂

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -494,7 +491,7 @@ class RelatedDocumentTableData
             public string SchemaName { get; set; }
             public string IdColumnName { get; set; }
             public string RelatedDocumentIdColumnName { get; set; }
-            public (string parentIdVariable, string relatedDocumentId, string relatedTableName)[] Related { get; set; }
+            public (string parentId, string relatedDocumentId, string relatedTableName)[] Related { get; set; }
```
flin-8, line 494: I need the actual value of the FK record, not its variable to insert it into the table valued parameter

source/Nevermore/Util/DataModificationQueryBuilder.cs
flin-8: @pawelpabich code changes are in, I am taking another look at the existing tests because I think the assertions may be too relaxed

flin-8: Also will do some quick benchmarking

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -487,14 +562,18 @@ into g
...
+        }
 
         class RelatedDocumentTableData
         {
             public string TableName { get; set; }
             public string SchemaName { get; set; }
             public string IdColumnName { get; set; }
             public string RelatedDocumentIdColumnName { get; set; }
-            public (string parentIdVariable, string relatedDocumentId, string relatedTableName)[] Related { get; set; }
+            public (string parentIdVariable, object parentId, string relatedDocumentId, string relatedTableName)[] Related { get; set; }
```
flin-8, line 576: We require the actual document ID when populating a table valued parameter

source/Nevermore.Benchmarks/Model/Order.cs
```diff
@@ -0,0 +1,25 @@
...
+        public Order(IEnumerable<(string, Type)> relatedDocuments)
+        {
+            RelatedDocuments = relatedDocuments?.ToArray();
+        }
+
+        public string Id { get; set; }
+        public string Name { get; set; }
+        public decimal Price { get; set; }
+
+        public IEnumerable<(string, Type)> RelatedDocuments { get; }
```
pawelpabich, line 23: The code I can find in OctopusServer has this property marked as `  [JsonIgnore]`. Should we do the same here?

source/Nevermore/UpgradeScripts/Script0003-Nevermore-RelatedDocumentType.sql
```diff
@@ -0,0 +1,2 @@
+﻿IF NOT EXISTS (SELECT NULL FROM sys.table_types WHERE name = 'RelatedDocumentTableValuedParameter')
```
flin-8, line 1: Made the upgrade scripts safe to re-run and added `Nevermore` to the name, because we're going to start including them inside Server and Octofront

source/Nevermore/UpgradeScripts/Script0001-Nevermore-KeyAllocation.sql
```diff
@@ -1,16 +1,16 @@
-﻿IF NOT EXISTS (SELECT NULL FROM sys.tables WHERE name = 'KeyAllocation')
-	CREATE TABLE KeyAllocation
+﻿IF NOT EXISTS (SELECT NULL FROM sys.tables WHERE name = 'KeyAllocation' AND SCHEMA_NAME(schema_id) = 'dbo')
```
flin-8, line 1: We should check the schema here too

source/Nevermore/IRelationalStoreConfiguration.cs
```diff
@@ -65,5 +65,10 @@ public interface IRelationalStoreConfiguration
         /// resulting in duplicate query plans being created.
         /// </summary>
         public bool DetectQueryPlanThrashing { get; set; }
+
+        /// <summary>
+        /// Gets or sets whether to support a larger number of related documents (currently Nevermore errors at 1000) by using table valued parameters
+        /// </summary>
+        public bool SupportLargeNumberOfRelatedDocuments { get; set; }
```
flin-8, line 72: Feature switch

source/Nevermore/UpgradeScripts/Script0001-Nevermore-KeyAllocation.sql
```diff
@@ -1,16 +1,16 @@
...
+	CREATE TABLE dbo.KeyAllocation
```
pawelpabich, line 2: I don't know enough about SQL schemas but is it possible that the schema was omitted on purpose?

flin-8: Server's version has `dbo`

https://github.com/OctopusDeploy/OctopusDeploy/blob/dd470d3ee864bf8b31683d2bb9b1e70b64e60ede/source/Octopus.Core/UpgradeScripts/Script0005%20-%20Key%20allocation.sql#L16

The fact that `ParameterList` has `dbo` prefixed makes me think that this was an omission

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -456,13 +464,80 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
...
+                sb.AppendLine($"INSERT INTO [{data.SchemaName}].[{data.TableName}] ([{data.IdColumnName}], [{data.IdTableColumnName}], [{data.RelatedDocumentIdColumnName}], [{data.RelatedDocumentTableColumnName}])");
+                sb.AppendLine($"SELECT [{data.IdColumnName}], [{data.IdTableColumnName}], [{data.RelatedDocumentIdColumnName}], [{data.RelatedDocumentTableColumnName}] FROM @{tableVariableName} t");
+                sb.AppendLine($"WHERE NOT EXISTS (SELECT null FROM [{data.SchemaName}].[{data.TableName}] r WHERE r.[{data.IdColumnName}] = @{IdVariableName} AND r.[{data.RelatedDocumentIdColumnName}] = t.[{data.RelatedDocumentIdColumnName}] )");
+            }
+        }
+
+        bool ShouldUseTableValuedParameter(RelatedDocumentTableData data)
+        {
+            //80 is roughly when using a table valued parameter is faster
+            return configuration.SupportLargeNumberOfRelatedDocuments && data.Related.Length >= 80;
```
pawelpabich, line 511: 👍

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -456,13 +464,80 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
...
+        }
+
+        static TableValuedParameter CreateRelatedDocumentTableValuedParameter(DocumentMap mapping,
+            RelatedDocumentTableData data)
+        {
+            var idMetaData = new SqlMetaData(data.IdColumnName, SqlDbType.NVarChar, 400);
```
pawelpabich, line 517: 400 seems to be used quite a few places, is that the same 400? If so, should we encapsulate it in a const??

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -456,13 +464,80 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
...
+                for (var x = 0; x < data.Related.Length; x++)
+                    parameters.Add(relatedVariablePrefix + x, data.Related[x].relatedDocumentId);
+            }
+
+            void UseTableValuedParameter(RelatedDocumentTableData data)
+            {
+                var tableVariableName = $"{index}__relatedDocumentTableValuedParameter";
+                parameters.AddTable(tableVariableName, CreateRelatedDocumentTableValuedParameter(mapping, data));
+
+                sb.AppendLine($"DELETE FROM [{data.SchemaName}].[{data.TableName}] WHERE [{data.IdColumnName}] = @{IdVariableName}");
```
pawelpabich, line 498: These 6 lines seem to be the same no matter which approach is taken. Should we extract the sql and parameterize with the table name?

source/Nevermore.Benchmarks/RelatedDocumentBenchmark.cs
```diff
@@ -0,0 +1,150 @@
...
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Data;
+using System.Linq;
+using System.Threading;
+using BenchmarkDotNet.Attributes;
+using Nevermore.Benchmarks.Model;
+using Nevermore.Benchmarks.SetUp;
+
+namespace Nevermore.Benchmarks
```
pawelpabich, line 11: Good to see perf tests 👍

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -420,34 +448,20 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
             sb.AppendLine(statement);
             sb.AppendLine();
 
-            if (relatedDocumentData.Any(d => d.Related.Any()))
-                sb.AppendLine("DECLARE @references as TABLE (Reference nvarchar(400), ReferenceTable nvarchar(400))");
+            if (!ShouldUseTableValuedParameterForUpdate() && relatedDocumentData.Any(d => d.Related.Any()))
+                sb.AppendLine($"DECLARE @references as TABLE ([{RelatedDocumentTableValuedParameterRelatedDocumentIdColumn}] nvarchar(400), [{RelatedDocumentTableValuedParameterRelatedDocumentTableColumn}] nvarchar(400))");
```
flin-8, line 452: Renamed the columns in this temp table to be the same as our table valued parameter table type, so we can reuse the query

source/Nevermore.IntegrationTests/Model/OrderMap.cs
```diff
@@ -8,6 +7,7 @@ public class OrderMap : DocumentMap<Order>
         public OrderMap()
         {
             RelatedDocuments(o => o.RelatedDocuments);
+            RelatedDocuments(o => o.RelatedDocuments, tableName: "AnotherRelatedTable");
```
flin-8, line 10: We were missing tests for multiple related document tables

source/Nevermore/IRelationalStoreConfiguration.cs
pawelpabich: What's the re removing this flag?  Delete in a few weeks/months if there are no problems in Octofront?

source/Nevermore/Util/DataModificationQueryBuilder.cs
pawelpabich: I was thinking more about passing a list of column names to a template.  Imagine there is a bug somewhere between line 498 and 504. How obvious it is that that the same bug exists between lines 481 and 487?

pawelpabich (approved):
LGMT + comments

source/Nevermore/IRelationalStoreConfiguration.cs
flin-8: Yes - once we've got this enabled in Octofront and Server for a few weeks, then we can remove this

source/Nevermore/Util/DataModificationQueryBuilder.cs
flin-8: The line numbers look wrong, but did you mean something like this?

https://github.com/OctopusDeploy/Nevermore/pull/153/commits/acfcf87ffe82c86bd1e7942b699f725c8762fc52#diff-94be40892362f618a2307a0d89bf656db3bcf5ab10f60619ac87954a1800647aR498

PR #154, opened 2021-07-13
Remove lock from RelationalTransactionRegistry

johnsimons:
As part of some performance analysis done in Octopus Server, I noticed that once the concurrency story increases this lock has a significant impact on performance.

This PR replaces the List/lock with a ConcurrentDictionary


MarkSiedle (approved):
Looks good to me. We'll get this in and start a branch to include this in server

PR #155, opened 2021-07-16
Ignore certain types of interpolation for purposes of SQL injection

droyad:
This will ignore `nameof`, numbers and constants. They are very unlikely to represent SQLi vectors.

### DIFF
--- a/source/Nevermore.Analyzers/NevermoreSqlInjectionAnalyzer.cs
+++ b/source/Nevermore.Analyzers/NevermoreSqlInjectionAnalyzer.cs
@@ -160,6 +160,9 @@ namespace Nevermore.Analyzers
             if (symbolType == null)
                 return false;
 
+            if (symbolType.EnumUnderlyingType != null)
+                return true;
+
             switch (symbolType.SpecialType)
             {
                 case SpecialType.System_Enum:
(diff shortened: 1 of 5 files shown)


source/Nevermore.Analyzers/NevermoreSqlInjectionAnalyzer.cs
```diff
@@ -95,13 +103,103 @@ static Location CheckForSqlInjection(ExpressionSyntax expression, SemanticModel
...
+                case SpecialType.System_Int16:
+                case SpecialType.System_UInt16:
+                case SpecialType.System_Int32:
+                case SpecialType.System_UInt32:
+                case SpecialType.System_Int64:
+                case SpecialType.System_UInt64:
+                case SpecialType.System_Decimal:
+                case SpecialType.System_Single:
+                case SpecialType.System_Double:
+                case SpecialType.System_DateTime:
```
tothegills, line 182: Worth testing `DateTime`?

tothegills (approved):
Neat PR, as I understand it adds things that are ok to interpolate in a SQL query because they are compile-time or incredibly unlikely to result in injection attack.

PR #156, opened 2021-07-18
Don't require key type for DELETE if you already have the document (similar to UPDATE)

flin-8:
This is reverting the signature change for requiring the key type on DELETE from https://github.com/OctopusDeploy/Nevermore/pull/149/files#diff-95a0a8cc5b6199fbc1e65c196e85c81d920a9d2a6112c86fc00a7598715f12cf

This makes it easier to call DELETE ([Slack conversation](https://octopusdeploy.slack.com/archives/C020HNMNT3L/p1626060225075400))

**Bumps major version**
`+semver:major`


slewis74 (approved):
Looks good 👍

PR #159, opened 2021-08-19
Fix type name in concurrency error message

adam-mccoy:
# Background

The error message produced in the case of a concurrency issue includes the name of the document type involved. This name is determined by using the generic type argument passed through to the `Update()` method. This type is not necessarily going to match the actual type of the document - it might be an interface implemented by the document, for example.

# Result

Take the document type name from the type stored on the mapping instead. This will be the actual type of the document.

## Before

```
Modification failed for 'IId`1' document with '1' Id because submitted data was out of date. Refresh the document and try again.
```

## After

```
Modification failed for 'Channel' document with '1' Id because submitted data was out of date. Refresh the document and try again.
```


uglybugger (approved):
Very nice 👍

PR #160, opened 2021-08-25
Bug: pass transaction names through to SQL Server

uglybugger:
Transaction name parameters are accepted by Nevermore but not propagated to SQL Server.

By the time an application becomes unresponsive it's very difficult to get a list of transactions from Nevermore, so we have to know to do it in advance.

With this change, we can just issue a `select * from sys.dm_tran_active_transactions` directly against SQL Server even if the application itself is unresponsive.


MarkSiedle (approved):
Small but very useful change, ty 👍

PR #164, opened 2021-09-03
Fix a bug around unique parameter reuse

APErebus:
There is a bug that occurs around parameter names when you re-use a query (via sub-query). This is a kinda esoteric bug and requires you to be re-adding the same where parameter on a query twice, however I did run into this issue when I was passing a base query to another method for re-use.

The error occurs as the `UniqueParameterNameGenerator` is unique per-transaction, not per-query. Also, if execute a query, a result of that is it removes the parameters from the unique name generator (after execution), which means if there are other queries using that parameter name, re-adding the parameter again causes a dictionary duplicate key exception.

The solution is to make the unique parameter name generator per-query, rather than per-transaction.

I did this by changing the transaction to have a builder func, rather than a single instance. Thus each QueryBuilder receives it's own instance of this `IUniqueParameterNameGenerator`, meaning each query (and subquery) maintains it's own list of unique parameter names.

### DIFF
--- a/source/Nevermore.IntegrationTests/Advanced/QueryReuseFixture.cs
+++ b/source/Nevermore.IntegrationTests/Advanced/QueryReuseFixture.cs
@@ -1,0 +1,27 @@
+﻿using System;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Nevermore.IntegrationTests.Model;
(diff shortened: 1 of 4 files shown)


APErebus:
Closing as I am no intending to fix this anymore and CPT is no longer using Nevermore

PR #166, opened 2021-09-22
Async Support on IRelatedDocumentStore

adam-mccoy:
## Background

Currently the `IRelatedDocumentStore` only has synchronous methods. These methods are called from the async methods on `IWriteQueryExecutor`, meaning they're likely to block threads.

This PR introduces async versions of these methods to avoid the blocking.

It also corrects an ambiguity problem in the `IRelatedDocumentStore` interface.

This is a breaking change to the interface, so will require a major bump.


MarkSiedle (approved):
Nice one. Looks like a sensible addition 👍

gupta-kartik:
Really nice!

PR #169, opened 2021-11-25
Column name cache and formatting select * to be select column names with json column last 

corey-underdown:
This change will remove the requirement for us to re-create tables when adding a type column to them by keeping a cached list of all the columns for a table when it's loaded for the first time, and replacing a `SELECT *` with a `SELECT column1, column2, column3, type, JSON` where the `JSON` column will always come last if it exists. 

This effects `Query` and the `Load` methods. For example, you can have a table that looks like:

Product
Id
JSON
Type

and doing `transaction.Query<Product>()` would construct a sql query like:

`SELECT Id,Type,JSON FROM Product` instead of `SELECT * FROM Product` like it used to. This allows `transation.Stream` to load in the Type column before the JSON column as it's been specified first, instead of using `*` and being restricted to the order of the table. 

It's possible to fall-back to the old way of `select *` by replacing the `TableColumnNameResolver` in the configuration with a new `ITableColumnNameResolver` that just returns `new[] {"*"}`. An example of this in Octopus Server: https://github.com/OctopusDeploy/OctopusDeploy/pull/10899


source/Nevermore/Querying/AST/ISelectColumns.cs
```diff
@@ -71,6 +71,23 @@ public class SelectAllSource : ISelectColumns
         public override string ToString() => GenerateSql();
     }
 
+    public class SelectAllJsonColumnLast : ISelectColumns
+    {
+        readonly string schema;
+        readonly string tableName;
+        readonly ITableColumnsCache tableColumnsCache;
+
+        public SelectAllJsonColumnLast(string schema, string tableName, ITableColumnsCache tableColumnsCache)
```
emilol, line 80: If we know the tablename at this point the schema should be guaranteed by now. Could we pass a readonly collection of resolved columns instead of the service?

source/Nevermore/TableColumnsCache.cs
```diff
@@ -0,0 +1,49 @@
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Nevermore
+{
+    public class TableColumnsCache : ITableColumnsCache
```
emilol, line 7: This is currently covering two concerns: resolving columns and caching them. Can we split these out and manage them separately?

source/Nevermore.IntegrationTests/RelationalStoreFixture.cs
```diff
@@ -57,6 +57,27 @@ public void ShouldPersistReferenceCollectionsToAllowLikeSearches()
             }
         }
 
+        [Test]
+        public void TableWithTypeAfterJsonShouldBeAbleToDeserialize()
```
emilol, line 61: Based on the test name, I would have expected to see a schema/map somewhere for this type that defines the Type column after the Json column. Where does that happen?

corey-underdown: The schema is generated for integration tests based on a list of document maps in `SchemaGenerator`. In there, I've made it so the json column is registered straight after Id, for all tables. This test then uses DodgyProduct, which is one of the few registered maps that actually have a Type specified for it.

source/Nevermore/Advanced/ReaderStrategies/Documents/DocumentReaderExpressionBuilder.cs
```diff
@@ -281,9 +281,6 @@ void AssertValidColumnOrdering()
             if (expectsType && typeIndex < 0)
                 throw Fail($"When querying the document '{name}', the '{map.TypeResolutionColumn.ColumnName}' column must always be included. Change the select clause to include the '{map.TypeResolutionColumn.ColumnName}' column.");
 
-            if (typeIndex >= 0 && expectsJson && typeIndex > jsonIndex)
-                throw Fail($"When querying the document '{name}', the 'Type' column must always appear before the 'JSON' column. Change the order in the SELECT clause, or if selecting '*', change the order of columns in the table.");
```
emilol, line 285: This still feels like a reasonable safety net, is it still necessary to remove this? Shouldn't we now be getting the columns ordered in such a way to avoid this perf issue?

corey-underdown: Good call - we should be keeping this for a case where someone might write their own query and not select in the right order.

source/Nevermore.IntegrationTests/Model/ProductSubtotalMap.cs
```diff
@@ -0,0 +1,15 @@
+using Nevermore.Mapping;
+
+namespace Nevermore.IntegrationTests.Model
+{
+    public class ProductSubtotalMap : DocumentMap<ProductSubtotal>
```
corey-underdown, line 5: The test using this previous never made a mapping for it, but it was sorta valid at the time since _RawSqlQuery_ method never actually loaded a map... It does now, and that feels valid to me since it's going to return a document type from it.

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
```diff
@@ -28,7 +29,13 @@ public void SetUp()
         
         ITableSourceQueryBuilder<TDocument> CreateQueryBuilder<TDocument>(string tableName, string schemaName = "dbo", string idColumnName = "Id") where TDocument : class
         {
-            return new TableSourceQueryBuilder<TDocument>(tableName, schemaName, idColumnName, transaction, tableAliasGenerator, uniqueParameterNameGenerator, new CommandParameterValues(), new Parameters(), new ParameterDefaults());
+            var columns = new[] {"*"};
```
corey-underdown, line 32: in this situation, if the test is passing `CreateQueryBuilder<object>` it's okay to use `*` as the column names. The test is checking that it's building the query correctly, and by saying the column name is `*` it still shows the builder used the value.

source/Nevermore/Advanced/SelectBuilderBase.cs
```diff
@@ -0,0 +1,192 @@
+﻿using System;
```
corey-underdown, line 1: These Select Builders have been given room to breath in their own files, instead of all being squished into the single file

emilol (approved):
LGTM, thank you for all of your hard work getting this one across the line 👍

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -403,11 +407,17 @@ await foreach (var item in StreamAsync<TDocument>(PrepareLoadMany<TDocument, TKe
...
+            
+            return new TableSourceQueryBuilder<TRecord>(map.TableName, schemaName, map.IdColumn?.ColumnName, this, tableAliasGenerator, ParameterNameGenerator, new CommandParameterValues(), new Parameters(), new ParameterDefaults());
         }
 
         public ISubquerySourceBuilder<TRecord> RawSqlQuery<TRecord>(string query) where TRecord : class
         {
+            //TODO: We're back to this problem... what do we send for column names? If we don't have column names, you can't join
+            // with re-ordered columns for you, but to have column names we need a map...
+            // if you really want to join, you'd probably do it in the query already, rather than using .Join so it's probably 
+            // just a limitation we accept now
```
emilol, line 420: is this comment still relevant?

source/Nevermore/CachingTableColumnNameResolver.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Linq;
+
+namespace Nevermore
+{
+    public class CachingTableColumnNameResolver : ITableColumnNameResolver
```
emilol, line 5: anything new that we don't _need_ to add to the public surface area of the library we should avoid doing so. Best to start with these being internal and change them later if necessary.

source/Nevermore/IReadQueryExecutor.cs
```diff
@@ -804,5 +804,8 @@ public interface IReadQueryExecutor
         /// <param name="cancellationToken">Token to use to cancel the command.</param>
         /// <returns>A data reader.</returns>
         [Pure] Task<DbDataReader> ExecuteReaderAsync(PreparedCommand preparedCommand, CancellationToken cancellationToken = default);
+
+        // TODO: work out a way to avoid needing to expose this
```
emilol, line 808: I think this comment is no longer relevant; we decided to go with exposing this since it's called from our querybuilders.

PR #171, opened 2021-12-14
Add support for JSON_VALUE queries

emilol:
Not all properties on a document need be mapped as columns, e.g. for `Product`:

```csharp
public class Product
{
    public string Id { get; set; }
    public string Name { get; set; }
    public decimal Price { get; set; }

    public ProductType Type { get; set; } = ProductType.Normal;
}
```

The nevermore mapping could be:

```csharp
public class ProductMap : DocumentMap<Product>
{
    public ProductMap()
    {
        Column(m => m.Name);
        TypeResolutionColumn(m => m.Type);
    }
}
```

This means that currently, if we try and query by anything other than `Name` or `Type` e.g:

```csharp
var products = t.Query<Product>()
    .Where(c => c.Price == 100)
    .ToList();
```

We get the following error:

```SQL
Error while executing SQL command in transaction 'NonParallelWorker': Invalid column name 'Price'.
The command being executed was:
SELECT Id,Name,Type,JSON
FROM [TestSchema].[Product]
WHERE ([Price] = @price)
ORDER BY [Id]
```

The `Price` field was queried as a column, instead of being parsed as part of the json payload like so:

```diff
SELECT Id,Name,Type,JSON
FROM [TestSchema].[Product]
-WHERE ([Price] = @price)
+WHERE (JSON_VALUE([JSON], '$.Price') = @price)
ORDER BY [Id]
```

### DIFF
--- a/source/Nevermore/Advanced/SelectBuilders/TableSelectBuilder.cs
+++ b/source/Nevermore/Advanced/SelectBuilders/TableSelectBuilder.cs
@@ -7,14 +7,14 @@ namespace Nevermore.Advanced.SelectBuilders
 {
     public class TableSelectBuilder : SelectBuilderBase<ITableSource>
     {
-        public TableSelectBuilder(ITableSource from, IColumn idColumn) 
+        public TableSelectBuilder(ITableSource from, IColumn idColumn)
             : this(from, idColumn, new List<IWhereClause>(), new List<GroupByField>(), new List<OrderByField>())
         {
         }
 
         TableSelectBuilder(ITableSource from, IColumn idColumn,
             List<IWhereClause> whereClauses, List<GroupByField> groupByClauses,
-            List<OrderByField> orderByClauses, ISelectColumns columnSelection = null, 
+            List<OrderByField> orderByClauses, ISelectColumns columnSelection = null,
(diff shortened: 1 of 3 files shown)


source/Nevermore/Advanced/SelectBuilders/TableSelectBuilder.cs
```diff
@@ -35,5 +37,17 @@ public override ISelectBuilder Clone()
         {
             return new TableSelectBuilder(From, IdColumn, new List<IWhereClause>(WhereClauses), new List<GroupByField>(GroupByClauses), new List<OrderByField>(OrderByClauses), ColumnSelection, RowSelection);
         }
+
+        public override void AddWhere(UnaryWhereParameter whereParams)
+        {
+            if (From.ColumnNames.Contains(whereParams.FieldName))
```
emilol: @adam-mccoy just a heads up on this one; `SelectAllColumnsTableResolver` will return the columns as `new [] { "*" }`.

It's not wired in by default since the most recent release (we now use `JsonLastTableColumnNameResolver` which _will_ return the column names) but this logic might need to be flipped to maintain backwards compat with the old resolver.

Issue #173, opened 2021-12-16
CachingTableColumnNameResolver only caches per transaction


droyad:
In Nevermore `16.2.4`  the `CachingTableColumnNameResolver` is re-created on every transaction ([change](https://github.com/OctopusDeploy/Nevermore/commit/3a897339fd6b461de0f0181212168a213d37c2cd)). This causes an extra query to be run on every transaction.

I ran up Octopus and clicked around a bit while recording the queries sent. I found that 45% of the queries were:
```
SELECT c.name\r\nFROM sys.tables AS t\r\nINNER JOIN sys.all_columns AS c ON c.object_id = t.object_id\r\nINNER JOIN sys.schemas AS s ON t.schema_id = s.schema_id\r\nWHERE t.name = @tableName AND s.name = @schemaName\r\nORDER BY (CASE WHEN c.name = 'JSON' THEN 1 ELSE 0 END) ASC, c.column_id"
```
or
```
SELECT c.name\r\nFROM sys.tables AS t\r\nINNER JOIN sys.all_columns AS c ON c.object_id = t.object_id\r\nINNER JOIN sys.schemas AS s ON t.schema_id = s.schema_id\r\nWHERE t.name = @tableName \r\nORDER BY (CASE WHEN c.name = 'JSON' THEN 1 ELSE 0 END) ASC, c.column_id"
 ```

**Comment by droyad** [comment]
This was [introduced into](https://github.com/OctopusDeploy/OctopusDeploy/commit/0f5ecda8f1a622f3bc654c548df452b854874545#diff-2c2481d51922f4bece815343ebd5a38b135a3dd750dc454f4727be10b385ce2cL47) Octopus in `2021.3.9503`

**Comment by corey-underdown** [comment]
Workaround: It's possible to replace the usage of CachingTableColumnNameResolver completely by overriding `TableColumnNameResolver` on the `RelationalStoreConfiguration`. By setting it to `SelectAllColumnsTableResolver`, it will mimic the same behaviour as before the change around `CachingTableColumnNameResolve` by doing a `Select *` again instead of `Select [column1], [column2], [json]`. 

eg.

```
var relationalStoreConfiguration = new RelationalStoreConfiguration(connectionString);
relationalStoreConfiguration.TableColumnNameResolver = _ => new SelectAllColumnsTableResolver();
```

emilol:
Fixed in #175

PR #175, opened 2021-12-16
Change TableColumnsCache scoping

emilol:
currently our cache is per transaction (relates to #173). 

This change elevates the cache so that it is shared between all transactions created from the same config.


corey-underdown (approved):
Oops, such a small thing to over look. Thanks for looking into it!

emilol:
Fixes #173

PR #177, opened 2022-01-18
Add query hint support

akirayamamoto:
This change adds support for Option query hints as shown below.

The Option clause has support for multiple query hints so I did the same here.

This would allow us to use `OPTIMIZE FOR UNKNOWN` to avoid parameter sniffing issues as described [here](https://techcommunity.microsoft.com/t5/sql-server-blog/optimize-for-unknown-8211-a-little-known-sql-server-2008-feature/ba-p/383498) and [here](https://blog.sqlauthority.com/2019/12/21/sql-server-parameter-sniffing-and-optimize-for-unknown/).

`OPTIMIZE FOR UNKNOWN` can be useful to solve this problem:
"An application that I work with presented me with an interesting dilemma; It wanted to utilize the benefits of plan reuse but the parameter values that the application initially sends to SQL Server are not representative of the values passed in the subsequent re-execution of the statement. SQL Server compiled and cached a ‘good’ plan for the first parameter values. Unfortunately, this had the unintended side effect of caching a poor execution plan for all subsequent parameter values."

### DIFF
--- a/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
+++ b/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Assent;
 using FluentAssertions;
-using Nevermore.Advanced;
 using Nevermore.Advanced.QueryBuilders;
 using Nevermore.Querying.AST;
 using NSubstitute;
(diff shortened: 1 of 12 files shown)


source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
```diff
@@ -1800,6 +1799,37 @@ public void ShouldGenerateSubqueryWhenThereAreNoCustomizations()
...
+                .Option("FORCE EXTERNALPUSHDOWN")
+                .DebugViewRawQuery();
+
+            const string expected = @"SELECT *
+FROM [dbo].[Orders]
+ORDER BY [Id]
+OPTION (OPTIMIZE FOR UNKNOWN, FORCE EXTERNALPUSHDOWN)";
+
+            actual.Should().Be(expected);
+        }
```
akirayamamoto, line 1832: This is the main goal of this PR.

LukeButters (approved):

MarkSiedle (approved):
Good tests 👍 This makes sense to me

PR #178, opened 2022-01-19
Fix query hint

akirayamamoto:
My [previous PR](https://github.com/OctopusDeploy/Nevermore/pull/177) would fail to generate a valid SQL query when calling `ToList(int skip, int take)`.

I created this PR to fix that.

Previously `ToList(int skip, int take)` would add the `OPTION` query hint inside the sub-query which is not a valid statement.

### DIFF
--- a/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
+++ b/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
@@ -1799,37 +1799,6 @@ ORDER BY [Id]";
 
             this.Assent(subquerySql);
         }
-
-        [Test]
-        public void ShouldGenerateOptionUnknown()
-        {
-            var actual = CreateQueryBuilder<object>("Orders")
-                .Option("OPTIMIZE FOR UNKNOWN")
-                .DebugViewRawQuery();
-
-            const string expected = @"SELECT *
-FROM [dbo].[Orders]
-ORDER BY [Id]
-OPTION (OPTIMIZE FOR UNKNOWN)";
-
-            actual.Should().Be(expected);
-        }
-
-        [Test]
-        public void ShouldGenerateOptionUnknownAndForceExternalPushDown()
-        {
-            var actual = CreateQueryBuilder<object>("Orders")
-                .Option("OPTIMIZE FOR UNKNOWN")
-                .Option("FORCE EXTERNALPUSHDOWN")
-                .DebugViewRawQuery();
-
-            const string expected = @"SELECT *
-FROM [dbo].[Orders]
-ORDER BY [Id]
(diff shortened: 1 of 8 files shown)


source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderStateFixture.cs
```diff
@@ -176,6 +199,70 @@ public void ShouldNotModifyQueryBuilderStateFromToListPageinatedWithCount()
...
+
+            queryBuilder.ParameterValues.Count.ShouldBeEquivalentTo(0);
+
+            queryBuilder.ToList();
+
+            LastExecutedQuery().ShouldBeEquivalentTo(@"SELECT *
+FROM [dbo].[Accounts]
+ORDER BY [Id]
+OPTION (OPTIMIZE FOR UNKNOWN)");
+        }
```
akirayamamoto, line 264: My [previous PR](https://github.com/OctopusDeploy/Nevermore/pull/177) would fail to generate a valid SQL query when calling `ToList(int skip, int take)`. I created this PR to fix that.

source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
```diff
@@ -1799,37 +1799,6 @@ public void ShouldGenerateSubqueryWhenThereAreNoCustomizations()
...
-                .Option("FORCE EXTERNALPUSHDOWN")
-                .DebugViewRawQuery();
-
-            const string expected = @"SELECT *
-FROM [dbo].[Orders]
-ORDER BY [Id]
-OPTION (OPTIMIZE FOR UNKNOWN, FORCE EXTERNALPUSHDOWN)";
-
-            actual.Should().Be(expected);
-        }
```
akirayamamoto, line 1832: These tests were replaced by those in `QueryBuilderStateFixture.cs`.

LukeButters (approved):

source/Nevermore/Advanced/QueryBuilder.cs
```diff
@@ -273,10 +275,16 @@ public async Task<int> CountAsync(CancellationToken cancellationToken = default)
...
             clonedSelectBuilder.AddColumnSelection(new SelectCountSource());
+            clonedSelectBuilder.AddOptions(optionClauses);
             var count = await readQueryExecutor.ExecuteScalarAsync<int>(clonedSelectBuilder.GenerateSelect().GenerateSql(), paramValues, RetriableOperation.Select, commandTimeout, cancellationToken);
             return count;
         }
 
+        IOption GetOption()
+        {
+            return new Option(optionClauses.Select(o => new OptionClause(o)).ToList());
+        }
```
LukeButters, line 286: Can this method be better named?

akirayamamoto: Oops, I just noticed I don't need that method anymore so I will just delete it.

PR #179, opened 2022-01-21
Add Hint and RawDebugView to Queryable

corey-underdown:
Adds `Hint` extension to `IQueryable` to be able to specify table hints in the query.
Adds `RawDebugQuery` extension to `IQueryable` to be able to see debug view the query like possible with using `Query`.

### DIFF
--- a/source/Nevermore/Advanced/Queryable/NevermoreQueryableExtensions.cs
+++ b/source/Nevermore/Advanced/Queryable/NevermoreQueryableExtensions.cs
@@ -53,12 +53,13 @@ namespace Nevermore.Advanced.Queryable
             if (source == null)
                 throw new ArgumentNullException(nameof(source));
             
-            var expression = Expression.Call(
-                null,
-                RawDebugViewInfo.MakeGenericMethod(typeof(TSource)),
-                source.Expression);
+            if (source.Provider is QueryProvider queryProvider)
+            {
+                var (command, _) = queryProvider.Translate(source.Expression);
+                return command.Statement;
+            }
 
-            return source.Provider.Execute<string>(expression);
+            throw new NotSupportedException();
         }
 
         public static async Task<int> CountAsync<TSource>(this IQueryable<TSource> source, CancellationToken cancellationToken = default)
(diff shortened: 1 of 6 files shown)


source/Nevermore/Advanced/Queryable/NevermoreQueryableExtensions.cs
```diff
@@ -32,6 +34,33 @@ public static IQueryable<TSource> WhereCustom<TSource>(this IQueryable<TSource>
...
+                throw new ArgumentNullException(nameof(hint));
+
+            return source.Provider.CreateQuery<TSource>(
+                Expression.Call(
+                    null,
+                    HintMethodInfo.MakeGenericMethod(typeof(TSource)),
+                    source.Expression, Expression.Constant(hint)));
+        }
+
+        public static string RawDebugView<TSource>(this IQueryable<TSource> source)
```
adam-mccoy, line 51: The debug functionality can be simplified by taking it out of the translation/execution process and instead allowing this extension to call `Translate()` on the IQueryable's Provider, assuming it is the right type. Something like:

```
if (source.Provider is QueryProvider queryProvider)
{
    var (command, _) = queryProvider.Translate(source.Expression);
    return command.Statement;
}

throw new NotSupportedException();
```

corey-underdown: Much simpler, thanks for that. I've removed the debug code from the query builder down too.

source/Nevermore/Advanced/Queryable/NevermoreQueryableExtensions.cs
```diff
@@ -11,6 +11,8 @@ namespace Nevermore.Advanced.Queryable
     public static class NevermoreQueryableExtensions
     {
         static readonly MethodInfo WhereCustomMethodInfo = new Func<IQueryable<object>, string, IQueryable<object>>(WhereCustom).GetMethodInfo().GetGenericMethodDefinition();
+        static readonly MethodInfo HintMethodInfo = new Func<IQueryable<object>, string, IQueryable<object>>(Hint).GetMethodInfo().GetGenericMethodDefinition();
+        static readonly MethodInfo RawDebugViewInfo = new Func<IQueryable<object>, string>(RawDebugView).GetMethodInfo().GetGenericMethodDefinition();
```
adam-mccoy, line 15: This can be removed now.

adam-mccoy (approved):
Looks good 🎉 Just one suggestion

PR #180, opened 2022-01-24
IQueryable - Handle polymorphic documents

adam-mccoy:
Fixes an issue where properties wouldn't map to columns for querying when the queried document type did not exactly match the type defined in the document map. This is a legitimate case for polymorphic document types.


emilol (approved):
Nice, a one-liner! Thanks for the quick turnaround, LGTM 👍

corey-underdown (approved):
Thanks for taking care of this Adam, change makes sense to me and looks good 👍

PR #182, opened 2022-01-28
Fixing issue with casting for types that inherit from base type with FirstOrDefaultAsync

corey-underdown:
When doing the following query:

`tx.Query<DodgyProduct>().Where(x => x.Name == "iphane").FirstOrDefaultAsync()`

We'd get an exception: `System.InvalidCastException: Object must implement IConvertable`. This was bubbling up from `(TResult)Convert.ChangeType(await FirstOrDefaultAsync(asyncStream, cancellationToken), typeof(TResult))` in the `QueryProvider.ExecuteAsync` which is only excercised through `IQueryable.GetFirstOrDefaultAsync` which we didn't have a test that hit. 

When it tries to convert from `DodgyProduct` to `Product`, it does a type comparison and `typeof(DodgyProduct) != typeof(Product)` so it was failing the convert type. The solution here is to not use ConvertType (which we believe was chosen as IAsyncEnumerable doesn't have a `.Cast` method like IEnumerable does, which is how `Execute` casts). Our work around here is horrible, but required. Since we don't have nullables enabled for Nevermore we can't convert `ExecuteAsync` to return `TResult?` which is _should_ be as it _does_ return null when no results are found. You can't return `(TResult) null` as it doesn't know how to convert a null to `TResult`, so the work around is hacky and yuck but works. 

We've also added sql timeout (-2) and row lock timeout (1222) to the command handler wrap exception so that when the sql exceptions are thrown, they will also be populated with open transactions at the time to help better diagnose the concurrency issues we're facing in Server.

### DIFF
--- a/source/Nevermore/Advanced/Queryable/QueryProvider.cs
+++ b/source/Nevermore/Advanced/Queryable/QueryProvider.cs
@@ -93,7 +93,7 @@ namespace Nevermore.Advanced.Queryable
                 {
                     if (typeof(TResult).IsValueType)
                     {
-                        throw new InvalidCastException("");
+                        throw new InvalidCastException("Null object cannot be converted to a value type");
                     }
 
                     // TODO: This NEEDS to go away when we turn nullable on in Nevermore
--- a/source/Nevermore/Advanced/Queryable/QueryProvider.cs
+++ b/source/Nevermore/Advanced/Queryable/QueryProvider.cs
@@ -70,6 +70,7 @@ namespace Nevermore.Advanced.Queryable
                 .Invoke(queryExecutor, new object[] { command });
         }
 
+        
         public async Task<TResult> ExecuteAsync<TResult>(Expression expression, CancellationToken cancellationToken)
         {
             var (command, queryType) = Translate(expression);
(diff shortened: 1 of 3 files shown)


source/Nevermore/Advanced/Queryable/QueryProvider.cs
```diff
@@ -86,7 +87,27 @@ public async Task<TResult> ExecuteAsync<TResult>(Expression expression, Cancella
             {
                 var asyncStream = (IAsyncEnumerable<object>)GenericStreamAsyncMethod.MakeGenericMethod(expression.Type)
                     .Invoke(queryExecutor, new object[] { command, cancellationToken });
-                return (TResult)Convert.ChangeType(await FirstOrDefaultAsync(asyncStream, cancellationToken), typeof(TResult));
```
adam-mccoy, line 89: To address the issue it should be sufficient to just remove the call to `Convert.ChangeType()`. It's just a remnant of how this method used to work, and is now unnecessary.

source/Nevermore/Advanced/Queryable/QueryProvider.cs
```diff
@@ -86,7 +87,27 @@ public async Task<TResult> ExecuteAsync<TResult>(Expression expression, Cancella
...
+                var firstOrDefaultAsync = await FirstOrDefaultAsync(asyncStream, cancellationToken);
+
+                if (firstOrDefaultAsync is null)
+                {
+                    if (typeof(TResult).IsValueType)
+                    {
+                        throw new InvalidCastException("Null object cannot be converted to a value type");
```
adam-mccoy, line 96: This is actually not what we want. `FirstOrDefault` should work with value types. Currently it can't, because the implementation of `FirstOrDefaultAsync` below only works with reference types. This is fine for now, because projection is not yet implemented, so you can only query complete documents, which must be classes.

source/Nevermore/Advanced/Queryable/QueryProvider.cs
```diff
@@ -86,7 +87,27 @@ public async Task<TResult> ExecuteAsync<TResult>(Expression expression, Cancella
...
+                    // TODO: This NEEDS to go away when we turn nullable on in Nevermore
+                    // This method needs to be able to return null for instances like `FirstOrDefaultAsync` 
+                    object GetNull() => null;
+                    return (TResult) GetNull();
+                }
+                
+                if (!firstOrDefaultAsync.GetType().IsAssignableFrom(typeof(TResult)))
+                {
+                    throw new InvalidCastException($"Type {firstOrDefaultAsync.GetType()} is not castable to type {typeof(TResult)}");
+                }
```
adam-mccoy, line 108: This check is replicating what you get for free by just doing the type cast below.

adam-mccoy (commented):
Bit of context added. Most of the defensive type checking is unnecessary at this stage because these conditions can't actually be achieved. It should be enough to just remove the call to `Convert.ChangeType`.

P.S. I'm not here...
zoidburg-woop

MarkSiedle (approved):
👍 Thank you. This looks like a good surgical change and a useful TODO for later.

PR #183, opened 2022-02-17
Add retry for BeginTransaction

paulegradie:
Octopus Deploy dev teams have noticed transient connectivity issues when calling `BeginTransaction` in `ReacTransaction.cs`. 
We are also behind 2 versions of SqlClient that may have beneficial bug fixes or performance improvements..

The objective of this PR is to:

 - Bump the version of SqlClient
 - Wrap these calls with retry logic provided by the RetryManager

[Slack Thread](https://octopusdeploy.slack.com/archives/C02SY496BAM/p1644994741806619)

### Before

TODO: get these running (currently hitting errors)

### After
TODO: Same

### DIFF
--- a/source/Nevermore.Benchmarks/SetUp/IntegrationTestDatabase.cs
+++ b/source/Nevermore.Benchmarks/SetUp/IntegrationTestDatabase.cs
@@ -13,7 +13,7 @@ namespace Nevermore.Benchmarks.SetUp
             var username = Environment.GetEnvironmentVariable("NevermoreTestUsername");
             var password = Environment.GetEnvironmentVariable("NevermoreTestPassword");
             testDatabaseName = Environment.GetEnvironmentVariable("NevermoreBenchmarkDatabase") ?? "Nevermore-Benchmarks";
-            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}")
+            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}; Encrypt=False;")
             {
                 ApplicationName = testDatabaseName,
             };
--- a/source/Nevermore.IntegrationTests/SetUp/IntegrationTestDatabase.cs
+++ b/source/Nevermore.IntegrationTests/SetUp/IntegrationTestDatabase.cs
@@ -14,7 +14,7 @@ namespace Nevermore.IntegrationTests.SetUp
             var username = Environment.GetEnvironmentVariable("NevermoreTestUsername");
             var password = Environment.GetEnvironmentVariable("NevermoreTestPassword");
             testDatabaseName = Environment.GetEnvironmentVariable("NevermoreTestDatabase") ?? "Nevermore-IntegrationTests";
-            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}")
+            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}; Encrypt=False;")
             {
                 ApplicationName = testDatabaseName,
             };
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -80,7 +80,7 @@ namespace Nevermore.Advanced
         public void Open(IsolationLevel isolationLevel)
         {
             Open();
-            Transaction = connection!.BeginTransaction(isolationLevel, SqlServerTransactionName);
+            Transaction = connection!.BeginTransactionWithRetry(isolationLevel, SqlServerTransactionName);
         }
 
         public async Task OpenAsync(IsolationLevel isolationLevel)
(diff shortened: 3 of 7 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -89,7 +89,7 @@ public async Task OpenAsync(IsolationLevel isolationLevel)
 
             // We use the synchronous overload here even though there is an async one, because the BeginTransactionAsync calls
```
paulegradie, line 90: Doesn't seem like a async overload will be very helpful.

source/Nevermore/Nevermore.csproj
```diff
@@ -45,7 +45,7 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Diagnostics.Contracts" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
-    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.4" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="4.1.0" />
```
paulegradie, line 48: I'd recommend against using the sqlClient retry provider for now until we have a much more stable build environment. Currently, I'm only aware of this option for opening connections (not beginning transactions, but this advice applies to both in case that comes to light). Perhaps we don't want to make that change at all regardless.

source/Nevermore.IntegrationTests/SetUp/IntegrationTestDatabase.cs
```diff
@@ -14,7 +14,7 @@ public IntegrationTestDatabase()
             var username = Environment.GetEnvironmentVariable("NevermoreTestUsername");
             var password = Environment.GetEnvironmentVariable("NevermoreTestPassword");
             testDatabaseName = Environment.GetEnvironmentVariable("NevermoreTestDatabase") ?? "Nevermore-IntegrationTests";
-            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}")
+            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}; Encrypt=False;")
```
paulegradie, line 17: Bumping the SqlClient altered the default for encryption. Most devs probably don't have this enabled on their dev SQL Server instance, so we can set this to avoid forcing folks to have to set another env var, or change their server settings.

source/Nevermore.Benchmarks/SetUp/IntegrationTestDatabase.cs
```diff
@@ -13,7 +13,7 @@ public IntegrationTestDatabase()
             var username = Environment.GetEnvironmentVariable("NevermoreTestUsername");
             var password = Environment.GetEnvironmentVariable("NevermoreTestPassword");
             testDatabaseName = Environment.GetEnvironmentVariable("NevermoreBenchmarkDatabase") ?? "Nevermore-Benchmarks";
-            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}")
+            var builder = new SqlConnectionStringBuilder($"Server={sqlInstance};Database={testDatabaseName};{(username == null ? "Trusted_connection=true;" : string.Empty)}; Encrypt=False;")
```
paulegradie, line 16: See below

source/Nevermore.IntegrationTests/SetUp/IntegrationTestDatabase.cs
akirayamamoto: Will we need to add this on all connection strings that we have on Octopus?

source/Nevermore/Transient/TransactionExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class TransactionExtensions
+    {
+        public static DbTransaction BeginTransactionWithRetry(this SqlConnection connection, IsolationLevel isolationLevel, string sqlServerTransactionName)
+        {
+            return BeginTransactionWithRetry(connection, isolationLevel, sqlServerTransactionName, RetryManager.Instance.GetDefaultSqlTransactionRetryPolicy());
+        }
+
+        public static DbTransaction BeginTransactionWithRetry(this SqlConnection connection, IsolationLevel isolationLevel, string sqlServerTransactionName, RetryPolicy retryPolicy)
+        {
+            return (retryPolicy ?? RetryPolicy.NoRetry).LoggingRetries("Beginning Database Transaction").ExecuteAction(() => connection.BeginTransaction());
```
akirayamamoto, line 17:
```suggestion
            return (retryPolicy ?? RetryPolicy.NoRetry).LoggingRetries("Beginning Database Transaction").ExecuteAction(() => connection.BeginTransaction(isolationLevel, sqlServerTransactionName));
```
Don't we need to continue passing those arguments?

akirayamamoto (approved):
I left some comments in there. Thanks Paul!

droyad:
> Will we need to add this on all connection strings that we have on Octopus?
@akirayamamoto We might have to set it to `False` just like with [this setting](https://github.com/OctopusDeploy/Issues/issues/6899)

source/Nevermore.IntegrationTests/SetUp/IntegrationTestDatabase.cs
paulegradie: Yeah possibly / probably 👍 - good call

source/Nevermore/Transient/TransactionExtensions.cs
paulegradie: 🤦  Yes, haha of course. Thanks for catching this :D

PR #187, opened 2022-02-18
Re-enable the querying of views

droyad:
In Nevermore 15.0.1 (Jul 2021) and earlier it was possible to map and query a view. However with the change where it looks at the schema for what columns exist, it fails to add columns to the query (despite them being included in the mapping).

This re-enabled this functionality and allows for querying of views.

Let me know if there is a different preferred way to query views.


emilol (approved):
LGTM thank you for taking care of this 👍

PR #188, opened 2022-03-07
Change default TableColumnNameResolver back to Select All

corey-underdown:
Previously in https://github.com/OctopusDeploy/Nevermore/pull/169 we changed the default behaviour of Nevermore to use a JSON last column table name resolver, which means queries would look like:

`SELECT Name,Age,Type,JSON FROM Person` instead of `SELECT * FROM Person` to avoid the issue of having to have the JSON column last when using the Type column in a polymorphic document structure.

After some discussion https://octopusdeploy.slack.com/archives/C020HNMNT3L/p1643763058763589 (internal link), I proposed that having this as the default behaviour probably isn't correct https://octopusdeploy.slack.com/archives/C02KHED7RRU/p1643837352093699 (internal link), and we're going back to the previous behaviour, with us overriding that in Octopus Server where we needed this change. 

Since we no longer have json last or caching as our default, I've added some tests around their behaviour.

Possible concerns: We had an issue with this behaviour around scoping https://github.com/OctopusDeploy/Nevermore/pull/175 . It's possible we could fall back into this issue. We will need to be careful when re-implementing in Octopus Server. 

- [x] Update the docs (https://github.com/OctopusDeploy/Nevermore/wiki/Configuration#table-column-name-resolver) around the usage of changing the default behaviour of `TableColumnNameResolver`

### DIFF
--- a/source/Nevermore/RelationalStore.cs
+++ b/source/Nevermore/RelationalStore.cs
@@ -17,14 +17,12 @@ namespace Nevermore
     {
         readonly Lazy<RelationalTransactionRegistry> registry;
         readonly Lazy<KeyAllocator> keyAllocator;
-        readonly ITableColumnsCache tableColumnsCache;
 
         public RelationalStore(IRelationalStoreConfiguration configuration)
         {
             Configuration = configuration;
             registry = new Lazy<RelationalTransactionRegistry>(() => new RelationalTransactionRegistry(new SqlConnectionStringBuilder(configuration.ConnectionString)));
             keyAllocator = new Lazy<KeyAllocator>(() => new KeyAllocator(this, configuration.KeyBlockSize));
-            tableColumnsCache = new TableColumnsCache();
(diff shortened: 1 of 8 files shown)


source/Nevermore/RelationalStore.cs
```diff
@@ -17,14 +17,12 @@ public class RelationalStore : IRelationalStore
     {
         readonly Lazy<RelationalTransactionRegistry> registry;
         readonly Lazy<KeyAllocator> keyAllocator;
-        readonly ITableColumnsCache tableColumnsCache;
```
corey-underdown, line 20: This wasn't ever being used, leftover artifact from when we were spiking how this was going to work

emilol (approved):
LGTM. I'd suggest even though this change is wire-compatible, we should look at bumping the major because of the change in behaviour and including a release note. What do you think?

Issue #83, opened 2019-09-23
Inherited properties aren't excluded from JSON


flin-8:
If you declare a child class that inherits some properties from its parent, the property won't be excluded from the JSON even when you map the property in the corresponding `DocumentMap`.

For example, see the following class hierarchy
```    public abstract class ParentClass :IDocument
    {
        public string Id { get; set; }
        public string Name { get; set; }
        public string ParentProperty { get; set; }
    }

    public class ChildClass : ParentClass
    {
        public string ChildProperty { get; set; }
    }
```

The `DocumentMap` for the child class looks like:
```    public class ChildClassMap : DocumentMap<ChildClass>
    {
        public ChildClassMap()
        {
            Column(m => m.ParentProperty);
            Column(m => m.ChildProperty);
        }
    }
```

You would expect both `ParentProperty` and `ChildProperty` to be excluded from the JSON. Only `ChildProperty` is excluded however, because in [`RelationalJsonContractResolver`](https://github.com/OctopusDeploy/Nevermore/blob/b65499654ada3aff0ea0b8b916bcb5d5a69984e7/source/Nevermore/RelationalJsonContractResolver.cs#L21), the `DocumentMap` is null because it's looking up the `DocumentMap` by the parent type `ParentClass`. 

Using the `ReflectedType` property doesn't help since the way `Newtonsoft` gets the properties [ensures it always reads the property from its declaring type](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/ReflectionUtils.cs#L951).

I'm not sure if there's an easy way around this other than making the user put `[JsonIgnore]` on the properties or making them abstract on the base class (so the child can redeclare them).

PR #193, opened 2022-03-24
Fix JSON Contract Resolver to support columns from base types

adam-mccoy:
# Background

The `RelationalJsonContractResolver` is responsible for determining which properties should be serialized in the JSON blob. It does this by using the declaring type of the property to fetch the document map and see whether it's mapped to a column, in which case it should be excluded.

This doesn't work if the property is declared by a base type that doesn't have a document map. In this case, if the derived type maps the property to a column, the property will be written to the JSON blob and the column.

# Result

Fix the resolver to always use the type for the contract being built to find the document map.


akirayamamoto (approved):
These changes make sense to me.

adam-mccoy:
Fixes: https://github.com/OctopusDeploy/Nevermore/issues/83

PR #194, opened 2022-04-12
Added transaction names

akirayamamoto:
One day we want transaction names to be mandatory. This PR gets us one step closer, at least on this repository.


gupta-kartik (approved):
Very tidy 🙇🏽

PR #198, opened 2022-05-03
Fix bug where we accept any SemaphoreSlim instead of a DeadlockAwareLock

uglybugger:
This bug was noticed by @pawelpabich [here](https://github.com/OctopusDeploy/Nevermore/pull/192#pullrequestreview-957294089).

It's currently a non-functional bug as it _just so happens_ that the type we're passing in is actually a `DeadlockAwareLock` but things will go sneakily bad if we pass in a bare `SemaphoreSlim` at any later date.

We'd initially intended to remove the inheritance of `SemaphoreSlim` from `DeadlockAwareLock` but we get too many nice extensions on `SemaphoreSlim` from `Nito.AsyncEx` which we'd have to reimplement so it actually looks safer to keep the inheritance but have the compiler make it impossible to pass an incorrect type around.

### DIFF
--- a/source/Nevermore/Advanced/ThreadSafeAsyncEnumerable.cs
+++ b/source/Nevermore/Advanced/ThreadSafeAsyncEnumerable.cs
@@ -9,21 +9,21 @@ namespace Nevermore.Advanced
     public class ThreadSafeAsyncEnumerable<T> : IAsyncEnumerable<T>
     {
         readonly Func<IAsyncEnumerable<T>> innerFunc;
-        readonly SemaphoreSlim semaphore;
+        readonly DeadlockAwareLock deadlockAwareLock;
 
-        public ThreadSafeAsyncEnumerable(IAsyncEnumerable<T> inner, SemaphoreSlim semaphore) : this(() => inner, semaphore)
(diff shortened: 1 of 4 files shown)


source/Nevermore/Advanced/ThreadSafeAsyncEnumerable.cs
```diff
@@ -9,21 +9,21 @@ namespace Nevermore.Advanced
     public class ThreadSafeAsyncEnumerable<T> : IAsyncEnumerable<T>
     {
         readonly Func<IAsyncEnumerable<T>> innerFunc;
-        readonly SemaphoreSlim semaphore;
+        readonly DeadlockAwareLock deadlockAwareLock;
 
-        public ThreadSafeAsyncEnumerable(IAsyncEnumerable<T> inner, SemaphoreSlim semaphore) : this(() => inner, semaphore)
+        public ThreadSafeAsyncEnumerable(IAsyncEnumerable<T> inner, DeadlockAwareLock deadlockAwareLock) : this(() => inner, deadlockAwareLock)
```
uglybugger, line 14: This is the actual bug. We were previously passing a `DeadlockAwareLock` here already so there should be no functional difference but we did set a trap for our future selves which we've hopefully now removed.

pawelpabich (approved):
LGTM

PR #200, opened 2022-06-06
Robe/cte count

zentron:
# Background
The `ToListWithCount()` operation is used in many different locations within Octopus Server. This operation as it is currently written results in 2 different operations, a paged query and a count query. Although the query is _largely_ the same for both, this is still 2 operations to the DBMS. 

## Solution
From some simple testing, converting the 2 queries into a single query via a CTE results in an improvement in test databases. This CTE includes the underlying data, paging, and crucially, another column which contains the _same_ count value on every record. This total count value is provided by a `CROSS APPLY` subquery with a `COUNT(*)` statement. In testing, including the count as a subquery directly in the `SELECT` statemen appeared to provide similar stats.

## Key Changes
- For testing and evaluation purposes, an [environment variable](https://github.com/OctopusDeploy/Nevermore/pull/200/files#diff-73f84433172c2736b07de81ebdaa986bee229c0be90da4f760b16bdc3eaee465R9) `NEVERMORE__UseCteBasedListWithCount` is added to allow opting into the new algorithm.
- New `ToListWithCount()` [operation](https://github.com/OctopusDeploy/Nevermore/pull/200/files#diff-5e2cb95109160b57b6e3671f113274fb13cea2fc6333f5cb55c6a76f62a49570R564) has been updated to [build](https://github.com/OctopusDeploy/Nevermore/pull/200/files#diff-5e2cb95109160b57b6e3671f113274fb13cea2fc6333f5cb55c6a76f62a49570R420) a CTE query
- ThreadSafe support extended to `StreamAsync` when [custom mapper is used](https://github.com/OctopusDeploy/Nevermore/pull/200/files#diff-cb574f356346d8539d664d75fd6253faf8ab060fb248df62fd320b0f5842214fR501).

## Take Note ⚠️
As noted, this change makes use of an environment-variable provided feature flag to opt into the new behavior. This will be turned on for some of our larger customers so we can assess if this actually provides better performance. Lower data sets result in almost like-for-like performance characteristics between the CTE and 2-query approach however there appeared to be a 50% improvement when working with larger datasets. Once we have confirmed the improvements we will remove the flag in the next few months. At that point we may also elevate CTE operations up as a more standard operation.

### DIFF
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -498,15 +498,20 @@ namespace Nevermore.Advanced
             }
         }
 
-        public async IAsyncEnumerable<TResult> StreamAsync<TResult>(string query, CommandParameterValues args, Func<IProjectionMapper, TResult> projectionMapper, TimeSpan? commandTimeout = null, [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        public IAsyncEnumerable<TResult> StreamAsync<TResult>(string query, CommandParameterValues args, Func<IProjectionMapper, TResult> projectionMapper, TimeSpan? commandTimeout = null, CancellationToken cancellationToken = default)
         {
-            var command = new PreparedCommand(query, args, RetriableOperation.Select, commandBehavior: CommandBehavior.Default, commandTimeout: commandTimeout);
-            await using var reader = await ExecuteReaderAsync(command, cancellationToken);
-            var mapper = new ProjectionMapper(command, reader, configuration.ReaderStrategies);
-            while (await reader.ReadAsync(cancellationToken))
+            async IAsyncEnumerable<TResult> Execute()
             {
-                yield return projectionMapper(mapper);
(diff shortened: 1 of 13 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -498,15 +498,20 @@ public IEnumerable<TResult> Stream<TResult>(string query, CommandParameterValues
...
+                var command = new PreparedCommand(query, args, RetriableOperation.Select, commandBehavior: CommandBehavior.Default, commandTimeout: commandTimeout);
+                await using var reader = await ExecuteReaderAsync(command, cancellationToken);
+                var mapper = new ProjectionMapper(command, reader, configuration.ReaderStrategies);
+                while (await reader.ReadAsync(cancellationToken))
+                {
+                    yield return projectionMapper(mapper);
+                }
             }
+
+            return new ThreadSafeAsyncEnumerable<TResult>(Execute, DeadlockAwareLock);
```
zentron, line 514: This method was missing the thread safe changes introduced in
https://github.com/OctopusDeploy/Nevermore/pull/192

uglybugger (approved):
This looks really tidy. Nice work 👍

Issue #201, opened 2022-06-10
DistinctAsync generates invalid SQL


adam-mccoy:
Using the `DistinctAsync()` method of the query builder will include the default `ORDER BY` clause. This can result in invalid SQL being generated if the columns in the `ORDER BY` clause do not appear in the select list. The following SQL error will be received:

```
SQL Error 145 - ORDER BY items must appear in the select list if SELECT DISTINCT is specified.
```

PR #202, opened 2022-06-10
Fix SQL generation bug when using DistinctAsync

adam-mccoy:
# Background

The `DistinctAsync()` method of the query builder does not exclude the default ORDER BY clause, like its synchronous counterpart.

Fixes https://github.com/OctopusDeploy/Nevermore/issues/201


skermajo (approved):
Nice find. 👍

PR #204, opened 2022-06-19
Support `CROSS JOIN`

zentron:
Adds support for `CROSS` join. This join is slightly different to standard LEFT or RIGHT joins in that there are no join clauses, the result is a cartesian product of the two sets.


adam-mccoy:
I'm curious as to why `CROSS APPLY` is used instead of `CROSS JOIN`. From memory, you'd normally use `CROSS APPLY` to join on the result of a function call. I'm not sure if it makes any difference here, just wondering if it was a deliberate decision.

zentron:
> I'm curious as to why `CROSS APPLY` is used instead of `CROSS JOIN`
Fair question. I think depending on the usage they could end up functionally equivalent. I suppose it might be a good question of why not support both?

The `CROSS APPLY` doesn't necessarily need to be a function, its just a "join" where the inner query can take a value coming from the outer source. It probably is interpreted better as a function as a result.

I can see it might be confusing conflating a JOIN with an APPLY so Ill convert this specifically to a `CROSS JOIN` to make it clearer.

adam-mccoy (approved):
Nice. LGTM 👍

PR #206, opened 2022-07-07
cover ToListWithCount where zero results are returned

zentron:
Fix for a bug in the new single-query-ToListWithCount operation where no results are returned. Since the total count is "piggy-backed" on the result set, if result set is returned then we miss out on the count. 

To work around this we fallback to just calling the count instead. This may result in performing 2 calls as we were originally trying to avoid, however this should be edge-case and no worse than what it already was before.


MarkSiedle (approved):
Good test coverage. Looks reasonable to me 👍

PR #207, opened 2022-07-15
Ensure a schema is passed to column name resolvers even if one isn't specified in a document map

MissedTheMark:
# Background #
If two tables with the same name exist in different schemas, and a schema name isn't provided for related document mappings, the `ReadTransaction` will pass a null schema name to any column resolvers.
In the case of `JsonLastTableColumnNameResolver`, this means columns from _both_ tables will be resolved💥 

This fix just changes `mapping.SchemaName` to `configuration.GetSchemaNameOrDefault(mapping)` in `ReadTransaction.PrepareLoad(..)` and `ReadTransaction.PrepareLoadMany(..)`.

## Notes ##
This issue could also be fixed by changing `JsonLastTableColumnNameResolver`, however I decided against that, for a couple of reasons:
- We have to get the schema from the map in `ReadTransaction` anyway, so it would be odd to get the schema there, then replace it in column name resolver if it's null.
- The `.Query()` method has to do (and already does) the same thing in this same class.
- And finally, if we do it in `JsonLastTableColumnNameResolver`, other column name resolvers could potentially have the same issue. 

Ideally testing for this would be done as unit tests, but given the design of the `ReadTransaction` class and its dependencies, we'd end up with a million mocks.

### DIFF
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -628,7 +628,7 @@ namespace Nevermore.Advanced
             if (mapping.IdColumn.Type != typeof(TKey))
                 throw new ArgumentException($"Provided Id of type '{id?.GetType().FullName}' does not match configured type of '{mapping.IdColumn?.Type.FullName}'.");
 
-            var columnNames = GetColumnNames(mapping.SchemaName, mapping.TableName);
+            var columnNames = GetColumnNames(configuration.GetSchemaNameOrDefault(mapping), mapping.TableName);
             var tableName = mapping.TableName;
             var args = new CommandParameterValues { { "Id", mapping.IdColumn.PrimaryKeyHandler.ConvertToPrimitiveValue(id) } };
 
@@ -642,7 +642,7 @@ namespace Nevermore.Advanced
             if (mapping.IdColumn?.Type != typeof(TKey))
                 throw new ArgumentException($"Provided Id of type '{typeof(TKey).FullName}' does not match configured type of '{mapping.IdColumn?.Type.FullName}'.");
 
-            var columnNames = GetColumnNames(mapping.SchemaName, mapping.TableName);
+            var columnNames = GetColumnNames(configuration.GetSchemaNameOrDefault(mapping), mapping.TableName);
             var param = new CommandParameterValues();
             param.AddTable("criteriaTable", idList.ToList(), configuration);
             var statement = $"SELECT s.{string.Join(',', columnNames)} FROM [{configuration.GetSchemaNameOrDefault(mapping)}].[{mapping.TableName}] s INNER JOIN @criteriaTable t on t.[ParameterValue] = s.[{mapping.IdColumn.ColumnName}] order by s.[{mapping.IdColumn.ColumnName}]";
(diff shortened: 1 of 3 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -642,14 +642,14 @@ protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand prepar
...
-            var columnNames = GetColumnNames(mapping.SchemaName, mapping.TableName);
+            var columnNames = GetColumnNames(configuration.GetSchemaNameOrDefault(mapping), mapping.TableName);
             var param = new CommandParameterValues();
             param.AddTable("criteriaTable", idList.ToList(), configuration);
             var statement = $"SELECT s.{string.Join(',', columnNames)} FROM [{configuration.GetSchemaNameOrDefault(mapping)}].[{mapping.TableName}] s INNER JOIN @criteriaTable t on t.[ParameterValue] = s.[{mapping.IdColumn.ColumnName}] order by s.[{mapping.IdColumn.ColumnName}]";
             return new PreparedCommand(statement, param, RetriableOperation.Select, mapping, commandBehavior: CommandBehavior.SingleResult | CommandBehavior.SequentialAccess);
         }
 
-        public string[] GetColumnNames(string? schemaName, string tableName)
+        public string[] GetColumnNames(string schemaName, string tableName)
```
droyad, line 652: Note: I wouldn't consider this a breaking change as it's not breaking at runtime -> nullability is only enforced by the compiler.

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -642,14 +642,14 @@ protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand prepar
             if (mapping.IdColumn?.Type != typeof(TKey))
                 throw new ArgumentException($"Provided Id of type '{typeof(TKey).FullName}' does not match configured type of '{mapping.IdColumn?.Type.FullName}'.");
 
-            var columnNames = GetColumnNames(mapping.SchemaName, mapping.TableName);
+            var columnNames = GetColumnNames(configuration.GetSchemaNameOrDefault(mapping), mapping.TableName);
```
droyad, line 645: Since it's used further down, I think it's worth pulling into a local variable `schema`

source/Nevermore/TableColumnNameResolvers/JsonLastTableColumnNameResolver.cs
```diff
@@ -14,10 +14,6 @@ public JsonLastTableColumnNameResolver(IRelationalStore store)
 
         public string[] GetColumnNames(string schemaName, string tableName)
         {
-            var schemaClause = string.IsNullOrEmpty(schemaName)
-                ? ""
-                : "AND s.name = @schemaName";
-
```
droyad, line 20: I checked all usages and schema shouldn't be null anymore (and if it's empty it's a configuration error).

droyad (approved):
Good pickup that we resolve the schema name like this elsewhere.

source/Nevermore/Advanced/ReadTransaction.cs
MissedTheMark: Good spot! I didn't even see that in the SQL below.

PR #208, opened 2022-08-12
Ensured rethrown SqlExceptions are being wrapped to preserve consumer exception handling assumptions

felixi-octopus:
# Background
The `WrapException` method rethrows an `Exception` when a `SqlException` is encountered. However, for certain types of `SqlExceptions`, it doesn't include the original `SqlException` as part of the rethrown `Exception`.

There are instances of consuming code that depends on the presence of this `SqlException` as an inner exception that break when more `Number`s are captured by this condition
```
if (ex is SqlException {Number: 1205 or 1222 or -2})
```
as it causes the raised `Exception` to exclude wrapping any inner exception.

# Before
No inner exception when deadlocks, timeouts and row lock timeouts are encountered. If other types of `SqlException`s are added in the same fashion, when those exceptions are raised, they will also stop containing inner exceptions.

# After
`SqlException` as an inner exception for all types of `SqlException`s

### DIFF
--- a/source/Nevermore/CommandExecutor.cs
+++ b/source/Nevermore/CommandExecutor.cs
@@ -215,7 +215,7 @@ namespace Nevermore
                 builder.AppendLine(ex.Message);
                 builder.AppendLine("Current transactions: ");
                 transaction.WriteCurrentTransactions(builder);
-                throw new Exception(builder.ToString());
+                throw new Exception(builder.ToString(), ex);
             }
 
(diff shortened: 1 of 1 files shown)


skermajo (approved):
Thanks! Makes sense 👍 . The only reason I can think of it not being originally included would be payload size of the exception itself, but given these are `SqlException` they would be well behaved.

PR #213, opened 2022-09-06
Add batch processing for insert many transaction 

michaelongso:
It was identified that for `insert many` transaction, there is a good possibility on exceeding 2100 SQL parameters limit, which caused `SQL Exception`.

This update is to process transaction in batches to avoid this issue

### DIFF
--- a/source/Nevermore/Advanced/WriteTransaction.cs
+++ b/source/Nevermore/Advanced/WriteTransaction.cs
@@ -16,6 +16,9 @@ namespace Nevermore.Advanced
     [DebuggerDisplay("{ToString()}")]
     public class WriteTransaction : ReadTransaction, IRelationalTransaction
     {
+        // SQL throws exception when parameters exceeded this number.
+        const int AllowedSqlParametersCount = 2100;
+        
         readonly IRelationalStoreConfiguration configuration;
         readonly IKeyAllocator keyAllocator;
         readonly DataModificationQueryBuilder builder;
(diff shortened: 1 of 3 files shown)


source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -69,16 +72,23 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
...
+            foreach (var documentsBatch in documentList.BatchWithBlockSize(batchBlockSize))
+            {
+                var documentsToInsert = documentsBatch.ToArray();
+                var command = builder.PrepareInsert(documentsToInsert, options);
+                foreach (var document in documentsToInsert) configuration.Hooks.BeforeInsert(document, command.Mapping, this);
 
-            var outputs = ExecuteDataModification(command);
-            ApplyNewRowVersionsIfRequired(documentList, command.Mapping, outputs);
-            ApplyIdentityIdsIfRequired(documentList, command.Mapping, outputs);
+                var outputs = ExecuteDataModification(command);
```
michaelongso, line 85: Lots of noise on the diff. Essentially adding indentation from the introduction of foreach

source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -36,13 +36,20 @@ public DataModificationQueryBuilder(IRelationalStoreConfiguration configuration,
...
             options ??= InsertOptions.Default;
             var mapping = GetMapping(documents);
+            ValidateMapping(mapping, options);
+            
+            return GetDocumentParameters(m => keyAllocator(m), options.CustomAssignedId, documents, mapping, DataModification.Insert);
+        }
 
-            if (mapping.IdColumn?.Direction == ColumnDirection.FromDatabase && options.CustomAssignedId != null)
-                throw new InvalidOperationException($"{nameof(InsertOptions)}.{nameof(InsertOptions.CustomAssignedId)} is not supported for identity Id columns.");
+        public PreparedCommand PrepareInsert(IReadOnlyList<object> documents, InsertOptions options = null)
```
michaelongso, line 48: Bad diff. Line 52 is the only update from original `PrepareInsert`

uglybugger (approved):
This looks pretty sensible to me. Nice work 👍 

It's worth keeping in mind (do we have a Trello card somewhere) that the block size should be constant for each record type so it should be fine (and a modest performance improvement) to cache that.

michaelongso:
> 

I will find the trello if not create one

PR #217, opened 2022-09-08
Add SQL Error Code 10936 to list of transient error codes

bradleyjford:
We (Cloud Platform) are frequently hitting the following error/exception when executing our E2E tests which utilise a SQL Azure Elastic Pool. 

`SQL Error 10936 - Resource ID : %d. The request limit for the elastic pool is %d and has been reached.`

There is little information on this specific error code available although there is an open [PR](https://github.com/MicrosoftDocs/azure-docs/issues/54599) on the Azure Docs repo for this to be documented.

This PR adds this specific error code to the list of `SimpleTransientErrorCodes` 

[EF Core treats this as a transient error](https://github.com/dotnet/efcore/pull/20970) and as such I would expect this to be a safe addition to Nevermore.

It is worth noting that the list of transient error codes in EF Core (where these were originally sourced from based on code comments) now includes a number of additional codes not reflected in Nevermore. I'm happy to submit another PR to update this if this would be preferred?


bradleyjford:
Closing in favour of #224

PR #218, opened 2022-09-09
Add missing line-handling attribute

michaelongso:
It was identified that line-handling is causing issue with the SQL migration scripts


skermajo (approved):
This looks like a fun gotcha.

Issue #216, opened 2022-09-07
Prefix NuGet package with Octopus


skermajo:
Octopus Deploy has recently reserved the `Octopus` prefix on NuGet. To avoid dependency confusion when referencing multiple NuGet package sources, this prefix is being appended to owned packages. Once `Octopus.Nevermore` and `Octopus.Nevermore.Analyzers` have been published to NuGet, the existing non-prefixed [Nevermore](https://www.nuget.org/packages/Nevermore) and [Nevermore.Analyzer](https://www.nuget.org/packages/Nevermore.Analyzers) packages will be marked as deprecated.

Migration involves simply changing project package references from `Nevermore` to `Octopus.Nevermore`, and similar for Analyzers.

PR #219, opened 2022-09-12
Adding Octopus prefix to avoid references dependency confusion

michaelongso:
This work is to address [issue 216](https://github.com/OctopusDeploy/Nevermore/issues/216).

Updating packages to prepend `Octopus`, a reserved prefix on nuget.org


stevencl840 (approved):
LGTM

PR #220, opened 2022-09-19
Allow query generator to discriminate by type

adam-mccoy:
# Background

Nevermore is able to store multiple document types in a single table. However, when building queries it is unable to differentiate between the types. This results in the wrong result being generated in certain cases, such as counting documents of a certain type.

For example, a base type `Product` has two derived types, `ProductA` and `ProductB`. These are stored in the `[dbo].[Product]` table, with a `[Type]` column with either the value "ProductA" or "ProductB". When trying to count the number of `ProductB` documents like:

``` CSharp
transaction.Query<ProductB>().Count()
```

will generate the following SQL

``` SQL
SELECT COUNT(*) FROM [dbo].[Product]
```

This will count instances of both `ProductA` and `ProductB`.

Currently the filter must be added manually because type resolvers don't make available the value that should be used to discriminate between each possible type.

# Results

**BREAKING CHANGE:** Change the type resolver interface to both convert the discriminator value to the target type, as well as the reverse. This allows the query builders to generate the WHERE clause necessary to filter by type. This include queries generated both by the `IQueryBuilder` interface as well as the `IQueryable` provider.

In the above example, the following SQL would now be generated:

``` SQL
SELECT COUNT(*) FROM [dbo].[Product] WHERE [Type] = 'ProductB'
```

### DIFF
--- a/source/Nevermore.IntegrationTests/Model/ProductConverter.cs
+++ b/source/Nevermore.IntegrationTests/Model/ProductConverter.cs
@@ -5,7 +5,7 @@ namespace Nevermore.IntegrationTests.Model
 {
     public class ProductTypeResolver : IInstanceTypeResolver
     {
-        public Type Resolve(Type baseType, object typeColumnValue)
+        public Type ResolveTypeFromValue(Type baseType, object typeColumnValue)
         {
             if (typeof(Product).IsAssignableFrom(baseType) && typeColumnValue is ProductType productType)
             {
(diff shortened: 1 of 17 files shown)


source/Nevermore.IntegrationTests/Model/ProductConverter.cs
```diff
@@ -17,5 +17,16 @@ public Type Resolve(Type baseType, object typeColumnValue)
...
+        public object ResolveValueFromType(Type type)
+        {
+            if (typeof(Product).IsAssignableFrom(type))
+            {
+                if (type == typeof(DodgyProduct)) return ProductType.Dodgy;
+                if (type == typeof(SpecialProduct)) return ProductType.Special;
+            }
+
+            return null;
+        }
```
skermajo, line 30: Nevermore doesn't have nullability?

skermajo (approved):
Thanks - this is v.nice to have. 👍

adam-mccoy:
Switched back to a draft after testing in Octopus Server. The changes to `IQueryBuilder` won't fly because of how joins work. I'll have to rethink this a bit.

akirayamamoto:
Please update [this issue](https://github.com/OctopusDeploy/ResearchAndDevelopment/issues/553) when this fix reaches Octopus server.

adam-mccoy:
I've addressed the concerns I had with this one. @skermajo would you mind having another look please?

skermajo:
Thanks @adam-mccoy - nothing further from me. Everything looks good 👍 .

PR #222, opened 2022-10-04
Fix the code to log the transactions when the connection reached 90% of the pool size

akirayamamoto:
The logic to LogHighNumberOfTransactions before this change was `numberOfTransactions >= MaxPoolSize OR numberOfTransactions == 90% of MaxPoolSize`. This seems odd.

I changed the logic to `numberOfTransactions >= 90% of MaxPoolSize`.

### DIFF
--- a/source/Nevermore/Advanced/RelationalTransactionRegistry.cs
+++ b/source/Nevermore/Advanced/RelationalTransactionRegistry.cs
@@ -1,10 +1,10 @@
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using Microsoft.Data.SqlClient;
 using Nevermore.Diagnositcs;
+
 #if NETFRAMEWORK
 using System.Data.SqlClient;
 #else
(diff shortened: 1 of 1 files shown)


source/Nevermore/Advanced/RelationalTransactionRegistry.cs
```diff
@@ -69,16 +69,13 @@ string BuildHighNumberOfTransactionsMessage()
 
         public void WriteCurrentTransactions(StringBuilder sb)
         {
-            ReadTransaction[] copy;
-            lock (transactions)
```
akirayamamoto, line 73: Locking here is not necessary, the ConcurrentDictionary will grab a lock when reading the keys.

adam-mccoy:
I don't disagree with you, but could you elaborate on why you think this is odd, and why the proposed change makes more sense? 😄

akirayamamoto:
It is odd to me because the code would log the transactions when it hit exactly 90% of the max pool size. Anything over 90% and under the max pool size will not be logged. Then when the number of transactions is over the max pool size it will log again. There is a gap between 90% and 100% that won't log. With this change I am making it log when it hits anything equals or over 90%.

Logging when we hit the max pool size could be too late, it would be good to have more of these log messages with the list of transactions when there is pressure on the pool.

I noticed that the above is valid when debugging log level is enabled. Otherwise, it will log the transaction names only once with error log level when it reaches the max connection pool size. Currently, we don't log debug messages in the Cloud.

adam-mccoy:
Exceeding 90% is not inherently a problem, but it might be worth a warning. That's why we log the DEBUG message at 90%. We then start making lots of noise when we reach/exceed 100%, when we _really_ have a problem.

How might logging this repeatedly at above 90% help us practically? What does it enable us to do? Keeping in mind that generating this log message isn't free. It could even be a bit costly, depending on the number of transactions and how many commands have been executed in each.

akirayamamoto:
I see your point, there's not that much value in logging the transactions every time it is above 90% and yes, there is a cost to list these transactions. I am closing this PR.

PR #224, opened 2022-10-09
Updates list of SQL Server transient error codes to match those defined in EF Core

bradleyjford:
This is an alternative PR to #217.

We (Cloud Platform) are frequently hitting the following error/exception when executing our E2E tests which utilise a SQL Azure Elastic Pool.

`SQL Error 10936 - Resource ID : %d. The request limit for the elastic pool is %d and has been reached.`

While adding this error code to the list of transient codes in Nevermore, I noticed that the list is considerably different/outdated(?) when compared to that of EF Core - where this list was originally sourced.

This PR updates the list of transient error codes to match those [now defined in EF Core](https://github.com/dotnet/efcore/blob/3a428827470ed6806739715b7607b8e7752084d8/src/EFCore.SqlServer/Storage/Internal/SqlServerTransientExceptionDetector.cs)

### DIFF
--- a/source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
+++ b/source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
@@ -18,6 +18,7 @@ namespace Nevermore.Transient
         static readonly int[] SimpleTransientErrorCodes = {
             // Details https://docs.microsoft.com/en-us/azure/sql-database/sql-database-develop-error-messages#database-connection-errors-transient-errors-and-other-temporary-errors
             // Copied from https://github.com/aspnet/EntityFrameworkCore/blob/10e553acc2/src/EFCore.SqlServer/Storage/Internal/SqlServerTransientExceptionDetector.cs
+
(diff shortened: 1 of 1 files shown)


source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
```diff
@@ -103,7 +189,11 @@ sealed class SqlDatabaseTransientErrorDetectionStrategy : ITransientErrorDetecti
             64,
             // DBNETLIB Error Code: 20
             // The instance of SQL Server you attempted to connect to does not support encryption.
-            20
+            20,
+            // This exception can be thrown even if the operation completed successfully, so it's safer to let the application fail.
+            // DBNETLIB Error Code: -2
+            // Timeout expired. The timeout period elapsed prior to completion of the operation or the server is not responding. The statement has been terminated.
+            //-2,
```
bradleyjford, line 196: This is commented out in the EF Core sources

uglybugger (approved):
Thanks, @bradleyjford . This looks like a very sensible change 👍

PR #227, opened 2022-11-10
Add retries policy for dbdatareader.read calls

johnsimons:


droyad:
[Possibly relevant](https://docs.google.com/document/d/1WXsdhyswjvITJ1Zq1wiDcPnRoaZxp9dPWow9tzSpXF4/edit#heading=h.vrzfmvubamqj)

sahanocto:
I'm not sure if this approach of retries is safe, the [EF Core documentation](https://learn.microsoft.com/en-us/ef/ef6/fundamentals/connection-resiliency/retry-logic#streaming-queries-are-not-supported) has a good explanation of why it might not be:
>Streaming is not supported when a retrying execution strategy is registered. This limitation exists because the connection could drop part way through the results being returned. When this occurs, EF needs to re-run the entire query but has no reliable way of knowing which results have already been returned (data may have changed since the initial query was sent, results may come back in a different order, results may not have a unique identifier, etc.).

johnsimons:
@sahanocto what do you think if we were to just look for specific transient errors, eg 1205 deadlock?

sahanocto:
I don't know. It still feels unsafe to me. I imagine the server would have already thrown away the whole transaction and you'd have to retry from higher up.

gupta-kartik:
I agree with Sahan here and the documentation for EF also mentions not supporting retries while streaming ([link](https://learn.microsoft.com/en-us/ef/ef6/fundamentals/connection-resiliency/retry-logic#streaming-queries-are-not-supported)) 

>I imagine the server would have already thrown away the whole transaction and you'd have to retry from higher up.

This is why we switched Nevermore retries off across the board in Octopus Server. The transaction was abandoned by that point and we saw evidence of queries being retried without being in a transaction 😬. 

Core Platform Productivity is currently evaluating the possibility of using SqlClient's retry capabilities as an alternative. Recent versions of Microsoft's SqlClient library have retry functionality built-in. We are investigating if we can leverage this to add retries back to Nevermore that are safe to use within transactions? It doesn't appear that this is available out-of-the-box, but there are extension points that we might be able to make use of. ([shortcut story](https://app.shortcut.com/octopusdeploy/story/30139/can-we-make-use-of-sql-client-retries-since-removing-nevermore-retries))

PR #229, opened 2022-11-23
Prevent isolation mode from being reused

johnsimons:
We saw evidence that queries being executed as part of `KeyAllocator` were inheriting the KeyAllocator isolation mode of Serializable instead of ReadCommitted.

Nevermore KeyAllocator is putting the connection into Serializable here: https://github.com/OctopusDeploy/Nevermore/blob/a22029b02aaae5095fd45ee8138e6cc8610802aa/source/Nevermore/Mapping/KeyAllocator.cs#L85
Then the connection gets put back on the connection pool, grabbed up again when we do a BeginReadTransactionAsync, and because nothing has reset the isolation level the query ends up running in Serializable instead of Read Committed.

### DIFF
--- a/source/Nevermore.Benchmarks/NevermoreBenchmark.cs
+++ b/source/Nevermore.Benchmarks/NevermoreBenchmark.cs
@@ -92,15 +92,6 @@ namespace Nevermore.Benchmarks
         }
         
         [Benchmark]
-        public List<Customer> SelectWithoutTransaction()
-        {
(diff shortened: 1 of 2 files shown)


adam-mccoy (approved):
As discussed, these overloads should ultimately be removed entirely, since they don't offer any value. This change serves as a good interim step to address the core problem without introducing a breaking change.

PR #230, opened 2022-11-23
PoC to re-execute a transaction

akirayamamoto:
[sc-30139]


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -627,13 +658,17 @@ public Task<DbDataReader> ExecuteReaderAsync(string query, CommandParameterValue
         public DbDataReader ExecuteReader(PreparedCommand preparedCommand)
         {
             using var command = CreateCommand(preparedCommand);
-            return command.ExecuteReader();
+            var result = command.ExecuteReader();
+            ExecutedCommands.Add(preparedCommand);
```
akirayamamoto, line 662: I thought it is worth re-executing commands issued through the `ExecuteReader` and `ExecuteScalar` methods because there might be `INSERT` and `UPDATE` statements on them. For example:

```
UPDATE FROM table SET ... WHERE ...;
SELECT * FROM table;
```

source/MyNevermoreTest/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+trn.Update(student);
+
+// sleep and manually kill the connection
+
+student.Age -= 1;
+trn.Update(student);
+
+trn.Replay();
+
+trn.Commit();
```
akirayamamoto, line 41: I was using this project to test my code because the current integration test will not actually commit data. I believe this feature is part of the test framework.

PR #232, opened 2022-12-05
WIP Removed two overloads of BeginReadTransaction that did not allow you …

stevencl840:
…to set the isolation level


adam-mccoy (approved):
LGTM 👍 Thanks for this

Note this is a breaking change, at the binary level at least. You'll need to add `+semver: breaking` in your commit message when you merge.

stevencl840:
> LGTM 👍 Thanks for this
> 
> Note this is a breaking change, at the binary level at least. You'll need to add `+semver: breaking` in your commit message when you merge.

Thanks for the tip

PR #233, opened 2022-12-07
Make IAsyncQueryProvider public

borland:


adam-mccoy (approved):
Yep, it's public 😀 
Not a breaking change so no special version bump required.

PR #234, opened 2022-12-08
Allow us to use string.Contains with a string comparison enum, which will be ignored

stevencl840:
At present nevermore will throw a  NotSupportedException if a query makes use of string.Contains with a StringComparison enum value set. 

It is believe this argument should be ignored but not cause Nevermore to throw an exception. This would be consistent with how it handles StartsWith and EndsWith

### DIFF
--- a/source/Nevermore/Advanced/Queryable/QueryTranslator.cs
+++ b/source/Nevermore/Advanced/Queryable/QueryTranslator.cs
@@ -260,6 +260,11 @@ namespace Nevermore.Advanced.Queryable
                     var (fieldReference, _) = GetFieldReferenceAndType(right);
                     return sqlBuilder.CreateWhere(fieldReference, invert ? ArraySqlOperand.NotIn : ArraySqlOperand.In, values);
                 }
+                else if (left!.Type == typeof(string) &&  right.Type == typeof(StringComparison))
+                {
+                    return CreateStringMethodWhere(expression, invert);
+                }
+
             }
 
             throw new NotSupportedException();


source/Nevermore/Advanced/Queryable/QueryTranslator.cs
```diff
@@ -260,6 +260,11 @@ IWhereClause CreateMethodCallWhere(MethodCallExpression expression, bool invert
                     var (fieldReference, _) = GetFieldReferenceAndType(right);
                     return sqlBuilder.CreateWhere(fieldReference, invert ? ArraySqlOperand.NotIn : ArraySqlOperand.In, values);
                 }
+                else if (left!.Type == typeof(string) &&  right.Type == typeof(StringComparison))
+                {
+                    return CreateStringMethodWhere(expression, invert);
+                }
+
```
adam-mccoy, line 267: I think it'd make more sense to handle this case inside the `CreateStringMethodWhere()` method. That would mean dropping the check for only one argument, which is fine.

Allowing the overload but ignoring the parameter is likely to lead to more head-scratching rather than less. Any idea how difficult it'd be to properly support it?

adam-mccoy: > StartsWith and EndsWith already work ignoring case in Nevermore.

Does this mean you can specify a `StringComparison` argument and Nevermore will allow it but not use it? If that's the case, then that sounds like a bug. As a general rule, we want the query translator to throw up its hands if it comes across anything it doesn't know how to deal with properly.

stevencl840: Investigating alternatives

PR #235, opened 2022-12-14
Core/poc retry transaction2

stevencl840:
[sc-30139]


source/Nevermore/Util/DataModificationQueryBuilder.cs
```diff
@@ -342,7 +342,7 @@ CommandParameterValues GetDocumentParameters(Func<DocumentMap, object> allocateI
             switch (mapping.JsonStorageFormat)
             {
                 case JsonStorageFormat.TextOnly:
-                    result[$"{prefix}{JsonVariableName}"] = serializer.SerializeText(document, mapping);
+                    result[$"{prefix}{JsonVariableName}"] = serializer.SerializeText(document, mapping).ReadToEnd();
```
stevencl840, line 345: Can't rewind textreaders, so as a work around for the POC I just output the string. Will need to come up with something better if we go ahead with this

source/Nevermore/CommandExecutor.cs
```diff
@@ -258,7 +258,7 @@ void DisposeOfParameters()
                 // Parameters can contain streams and other disposable objects.
                 if (parameter.Value is IDisposable disposable)
                 {
-                    disposable.Dispose();
+                    //disposable.Dispose();
```
stevencl840, line 261: We can dispose of these at the transaction level once we are done. Just left it like this as it is only a PoC

source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
```diff
@@ -220,6 +220,7 @@ public bool IsTransient(Exception ex)
             }
 
             // Otherwise it could be another simple transient error
+            var result = sqlErrors.Select(x => x.Number).Intersect(SimpleTransientErrorCodes).Any();
```
stevencl840, line 223: Clean me

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -714,7 +767,9 @@ RetryPolicy GetRetryPolicy(RetriableOperation operation)
             if (operationsToRetry == RetriableOperation.None)
                 return RetryPolicy.NoRetry;
 
-            return (operationsToRetry & operation) != 0 ? RetryManager.Instance.GetDefaultSqlCommandRetryPolicy() : RetryPolicy.NoRetry;
+            //return (operationsToRetry & operation) != 0 ? RetryManager.Instance.GetDefaultSqlCommandRetryPolicy() : RetryPolicy.NoRetry;
+
+            return RetryManager.Instance.GetDefaultSqlCommandRetryPolicy();
```
stevencl840, line 772: We are not rolling back on an operation, instead a whole transaction so I short circuited this  for the purposed of POC

source/Nevermore/CommandExecutor.cs
```diff
@@ -42,13 +42,13 @@ public CommandExecutor(DbCommand command, PreparedCommand prepared, RetryPolicy
...
 
-        public int ExecuteNonQuery()
+        public int ExecuteNonQuery(Action replayClosure)
         {
             AssertSynchronousOperation();
             try
             {
                 AssertSynchronousOperation();
-                return command.ExecuteNonQueryWithRetry(retryPolicy);
+                return command.ExecuteNonQueryWithRetry(retryPolicy, replayClosure);
```
akirayamamoto, line 51: Interesting approach :)

akirayamamoto (commented):
Nice approach, it looks like this PoC will make use of the `GetRetryPolicy` method which means we will re-use the existing policy code as we discussed.

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -565,23 +644,47 @@ public int ExecuteNonQuery(string query, CommandParameterValues? args = null, Ti
...
 
         public int ExecuteNonQuery(PreparedCommand preparedCommand)
         {
             using var mutex = DeadlockAwareLock.Lock();
-            using var command = CreateCommand(preparedCommand);
-            return command.ExecuteNonQuery();
+            var retryPolicy = GetRetryPolicy(preparedCommand.Operation).LoggingRetries(nameof(ExecuteNonQuery));
+            var result = retryPolicy.ExecuteAction(() =>
+            {
+                if (connection?.State is ConnectionState.Closed)
```
stevencl840, line 658: Need to investigate if this is sufficient

akirayamamoto: I wonder if we can make use of `@@TRANCOUNT == 1`. Will we ever have multiple transactions open one on top of the other?

PR #241, opened 2023-01-06
ConfigureAwait(false) wherever we await a task

rosslovas:
The boolean passed to `ConfigureAwait` allows you to specify whether or not the callback after the `await` is forced to be invoked on the original `SynchronizationContext`. If unspecified, the default value is `true`. Under most circumstances this is not an issue, but for library code it's a good idea to specify `false` so the caller can decide.

This PR enables the [CA2007](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007) warning as an error to help guide us toward using `ConfigureAwait(false)`, and fixes all violations.

### DIFF
--- a/source/Nevermore/Advanced/DeadlockAwareLock.cs
+++ b/source/Nevermore/Advanced/DeadlockAwareLock.cs
@@ -31,7 +31,7 @@ namespace Nevermore.Advanced
         public new async Task WaitAsync(CancellationToken cancellationToken)
         {
             AssertNoDeadlock();
-            await base.WaitAsync(cancellationToken);
+            await base.WaitAsync(cancellationToken).ConfigureAwait(false);
             RecordLockAcquisition();
         }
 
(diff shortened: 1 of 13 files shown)


skermajo:
Given we're under .NET Core, do we have a `SynchronizationContext`?

rosslovas:
> Given we're under .NET Core, do we have a `SynchronizationContext`?

When testing with xUnit we do 🙂 [AsyncTestSyncContext](https://github.com/xunit/xunit/blob/77a3667654612e3ef7ba870f4e534c8ac7bebd13/src/xunit.v3.core/Sdk/AsyncTestSyncContext.cs) (there's also `MaxConcurrencySyncContext`, but we opted out of that recently. `AsyncTestSyncContext` can't be opted out of until xUnit v3.)

Regardless, this is a best-practice thing for any public library to be doing, so Nevermore should do it anyway.

adam-mccoy (approved):
LGTM 👍 Thanks for the effort.

PR #242, opened 2023-01-08
Fixed NOT IN not working with empty collection

### LINKED ISSUE #238: NOT IN operator doesn't work as expected when the target collection is empty
An example that works:

```csharp
var installedTemplatesLookup = new[] {"somevaluethatdoesntexist"};

var communityTemplates =
  await communityActionTemplateStore.Query().Where(t => !installedTemplatesLookup.Contains(t.Id)).ToListAsync(cancellationToken);
```
And this doesn't work:

```csharp
var installedTemplatesLookup = Array.Empty<string>();
            
var communityTemplates =
   await communityActionTemplateStore.Query().Where(t => !installedTemplatesLookup.Contains(t.Id)).ToListAsync(cancellationToken);
```

The expectation is that when the lookup is an empty collection the query simply returns the entire table.

APErebus:
Fixes #238 

The generated `WHERE` clause for an empty collection needed to change when using the `NOT IN` operator.

In that case, we want to include _all_ results, so a custom where of `1 = 1` is the way to go. When it's an `IN` operator, we do `1 = 0` as we shouldn't include _any_ results in that case

### DIFF
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -570,9 +570,9 @@ namespace Nevermore.IntegrationTests
 
             var testCustomers = new[]
             {
-                new Customer { FirstName = "Alice", LastName = "Apple", Roles = { "RoleA", "RoleB" }},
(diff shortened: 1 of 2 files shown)


adam-mccoy (approved):
LGTM 👍 Thanks for the fix.

PR #243, opened 2023-01-27
Retry pooled connection timeout

adam-mccoy:
## Background

If acquiring a SQL connection from the connection pool times out due to no connections being available, the operation will fail with an error. This class of error is currently not identified as a transient error, though other timeout errors are. This means that encountering this error would fail the operation immediately without retrying.

## Results

Add the exception raised by SqlClient when a pooled connect timeout occurs to the list of transient errors that could be retried. It is no different from other types of timeout errors, except that it is raised as an `InvalidOperationException` instead of a `SqlException`.

### DIFF
--- a/source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
+++ b/source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
@@ -201,11 +201,14 @@ namespace Nevermore.Transient
             return ex switch
             {
                 TimeoutException => true,
+                InvalidOperationException invalidOperationException => IsPooledConnectTimeout(invalidOperationException),
                 SqlException sqlException => IsTransientSqlException(sqlException),
                 _ => false
             };
         }
 
+        static bool IsPooledConnectTimeout(InvalidOperationException exception)
+             => exception.Message.Contains("The timeout period elapsed prior to obtaining a connection from the pool.");
 
         static bool IsTransientSqlException(SqlException exception)
         {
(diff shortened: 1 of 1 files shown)


source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
```diff
@@ -198,18 +198,27 @@ sealed class SqlDatabaseTransientErrorDetectionStrategy : ITransientErrorDetecti
...
+                InvalidOperationException invalidOperationException => IsPooledConnectTimeout(invalidOperationException),
+                SqlException sqlException => IsTransientSqlException(sqlException),
+                _ => false
+            };
+        }
 
-            var sqlException = ex as SqlException;
-            if (sqlException == null) return false;
+        static bool IsPooledConnectTimeout(InvalidOperationException exception)
+             => exception.Message.Contains("The timeout period elapsed prior to obtaining a connection from the pool.");
```
adam-mccoy, line 211: This is not ideal, but there is no error code or similar attached to the exception raised. Type and message string is all we have to go on ([source](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/Common/AdapterUtil.cs#L1141-L1142)).

YuKitsune (approved):
:shipit:

source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
tekerson: I am curious if this will break if the system language isn't English? and do we support that in the first place?

(I don't think it's a showstopper anyway, it will just go back to the current behaviour)

tekerson (approved):

source/Nevermore/Transient/SqlDatabaseTransientErrorDetectionStrategy.cs
adam-mccoy: Yeah, I considered this too. From what I can tell, SqlClient doesn't come with localised versions of the strings, only English.

PR #244, opened 2023-02-01
Update CODEOWNERS

adam-mccoy:
Update CODEOWNERS file to reflect new teams.


skermajo (approved):
This explains why I was still getting notifications :) Makes sense to me!

PR #246, opened 2023-02-03
Speed up tests

adam-mccoy:
## Background

A recent change made pooled connect timeouts a retriable error. An unintended consequence of this is now the test run takes a lot longer. This is due to the `RelationalTransactionRegristryFixture`, which deliberately exhausts the connection pool to validate some functionality. The test still works as intended, it just takes significantly longer now due to the retries being done.

## Result

Disable retries for the affected fixture. Configuring retries is done in global state, so it's possible this _may_ affect other test fixtures too. Ideally, retries would be part of the relational store configuration, but that's a problem for another day.

### Before

Test suite runtime: ~10 minutes

### After

Test suite runtime: ~4.5 minutes


rosslovas (approved):
Seems like a reasonable compromise, I agree that this alone probably doesn't justify the effort in changing the way retries are configured right now. Confirmed it speeds the affected tests up significantly both locally and in CI 👍

PR #247, opened 2023-02-08
Fix the project file reference on Nuke

akirayamamoto:
I could not get Nuke to build on my machine without this change.

`SourceDirectory` uses the full path and fixes the problem.

This is the command that was not working: `build.cmd --Target Restore`

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -42,7 +42,7 @@ class BuildNevermore : NukeBuild
         .Executes(() =>
     {
         DotNetRestore(_ => _
-            .SetProjectFile("source"));
+            .SetProjectFile(SourceDirectory));
     });
 
     Target Build => _ => _
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -1,5 +1,6 @@
 using Nuke.Common;
 using Nuke.Common.IO;
+using Nuke.Common.ProjectModel;
 using Nuke.Common.Tools.DotNet;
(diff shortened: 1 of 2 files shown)


build/Build.cs
```diff
@@ -42,7 +42,7 @@ class BuildNevermore : NukeBuild
         .Executes(() =>
     {
         DotNetRestore(_ => _
-            .SetProjectFile("source"));
+            .SetProjectFile(SourceDirectory));
```
stevencl840, line 45: The change makes sense to me, do we need feedback from other OS's?

akirayamamoto: It wouldn't hurt to try and build on Mac. I don't think it will be an issue since it works on Linux as we can see [here](https://build.octopushq.com/buildConfiguration/OctopusDeploy_LIbraries_Nevermore?branch=refs%2Fheads%2Fakira%2Ffix-nuke-build&buildTypeTab=overview&mode=builds).

@matt-richardson did you want to try and build this on your Mac?

akirayamamoto:
Heads up @stevencl840, I am now using source generated `Solution` like Matt suggested.

build/Build.cs
matt-richardson: Seems to run fine 👍

matt-richardson (approved):
looks :chefkiss: to me.

stevencl840 (approved):
Nice clean update

PR #248, opened 2023-02-14
Added lhs query alias for joins

benPearce1:
[sc-39691]

### DIFF
--- a/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
+++ b/source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
@@ -1887,6 +1887,51 @@ ORDER BY ALIAS_GENERATED_2.[Id]";
 
             actual.Should().BeEquivalentTo(expected);
         }
+        
+        [Test]
+        public void ShouldGenerateStaticJoinAliasNamesInJoin()
+        {
+            string actual = null;
+            CommandParameterValues parameters = null;
+            transaction.Stream<TodoItem>(Arg.Do<string>(s => actual = s),
+                Arg.Do<CommandParameterValues>(p => parameters = p));
+
+            var createdDate = DateTime.Now;
+            var joinDate = createdDate - TimeSpan.FromDays(1);
+            var sharedFieldName = "Date";
+
+            var orders = CreateQueryBuilder<object>("Orders")
+                .Where(sharedFieldName, UnarySqlOperand.Equal, createdDate);
+
+
+            var query = CreateQueryBuilder<TodoItem>("Customer")
+                .Where(sharedFieldName, UnarySqlOperand.Equal, joinDate)
+                .InnerJoin(orders.Subquery().Alias("o"), "j")
+                .On("Id", JoinOperand.Equal, "CustomerId");
+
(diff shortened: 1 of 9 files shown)


source/Nevermore.Tests/QueryBuilderFixture/QueryBuilderFixture.cs
```diff
@@ -1887,6 +1887,51 @@ public void ShouldGenerateUniqueParameterNamesInJoin()
 
             actual.Should().BeEquivalentTo(expected);
         }
+        
+        [Test]
+        public void ShouldGenerateStaticJoinAliasNamesInJoin()
```
benPearce1, line 1892: This test is a copy of `ShouldGenerateUniqueParameterNamesInJoin` above it, but with static aliases

source/Nevermore/Advanced/QueryBuilders/SubquerySourceBuilder.cs
```diff
@@ -25,7 +25,7 @@ protected override ISelectBuilder CreateSelectBuilder()
             return new SubquerySelectBuilder(AsSource());
         }
 
-        public override IJoinSourceQueryBuilder<TRecord> Join(IAliasedSelectSource source, JoinType joinType, CommandParameterValues parameterValues, Parameters parameters, ParameterDefaults parameterDefaults)
+        public override IJoinSourceQueryBuilder<TRecord> Join(IAliasedSelectSource source, JoinType joinType, CommandParameterValues parameterValues, Parameters parameters, ParameterDefaults parameterDefaults, string queryAlias = null)
```
benPearce1, line 28: Not sure of the use of the `Join` method here, the `JoinSourceQueryBuilder` purpose seems to be towards setting up the `ON` statement, so the function and the additional `queryAlias` here serve no purpose, unless I am missing something.

source/Nevermore/Advanced/QueryBuilders/JoinSourceQueryBuilder.cs
```diff
@@ -38,7 +38,7 @@ protected override ISelectBuilder CreateSelectBuilder()
             return new JoinSelectBuilder(joinedSource);
         }
 
-        public override IJoinSourceQueryBuilder<TRecord> Join(IAliasedSelectSource source, JoinType joinType, CommandParameterValues parameterValues, Parameters parameters, ParameterDefaults parameterDefaults)
+        public override IJoinSourceQueryBuilder<TRecord> Join(IAliasedSelectSource source, JoinType joinType, CommandParameterValues parameterValues, Parameters parameters, ParameterDefaults parameterDefaults, string queryAlias = null)
```
benPearce1, line 41: Not sure of the use of the `Join` method here, the `JoinSourceQueryBuilder` purpose seems to be towards setting up the `ON` statement, so the function and the additional `queryAlias` here serve no purpose, unless I am missing something.

tothegills: Yeah, this one looks a bit awkward, like the abstraction isn't quite right.

tothegills (approved):

PR #251, opened 2023-04-14
Add async enumeration support to INevermoreQueryable

rosslovas:
# Background

This PR adds the minimum requirements to properly support async enumeration over the results of a query via Nevermore's `IQueryable` implementation. It's backed by `StreamAsync` which already existed and works well, so not much had to be done.

It's possible to go further with this and add an `IAsyncEnumerable` layer on top which can provide `async foreach` support and nicer cancellation token ergonomics among other things, but that's left up to consumers.

# Notes

To confirm this works as intended (minimising the amount of information kept in memory while iterating) I ran the following code locally and used the debugger to verify the number of `Customer` instances on the heap did not grow.

```csharp
[Test]
public async Task TestingMemoryUsage()
{
    using var t = Store.BeginTransaction();

    foreach (var i in Enumerable.Range(0, 100))
    {
        t.Insert(new Customer { FirstName = $"Alice{i}", LastName = "Apple", Roles = { "Admin" } });
    }

    GC.Collect();

    var queryable = (INevermoreQueryable<Customer>)t.Queryable<Customer>();
    var enumerator = queryable.GetAsyncEnumerator(CancellationToken.None);
    while (await enumerator.MoveNextAsync())
    {
        var customer = enumerator.Current;

        // The number of Customer instances in memory should not grow here
        GC.Collect();
    }
}
```

Placing a breakpoint in the loop, I observed the number of `Customer` instances in memory never grew above 3 (looks like there's a bit of internal buffering going on but not much). For example, see the screenshot below in which there's only 3 in memory after iterating over 80+:

image

### DIFF
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1367,7 +1367,7 @@ namespace Nevermore.IntegrationTests
         }
 
         [Test]
-        public async Task AsyncForeach()
+        public async Task AsyncEnumeration()
         {
             using var t = Store.BeginTransaction();
 
@@ -1385,9 +1385,11 @@ namespace Nevermore.IntegrationTests
 
             var customers = new List<Customer>();
             var queryable = (INevermoreQueryable<Customer>)t.Queryable<Customer>().Where(c => c.FirstName.EndsWith("e"));
-            await foreach (var customer in queryable)
+
+            var enumerator = queryable.GetAsyncEnumerator(CancellationToken.None);
+            while (await enumerator.MoveNextAsync())
             {
-                customers.Add(customer);
+                customers.Add(enumerator.Current);
             }
 
             customers.Select(c => c.LastName).Should().BeEquivalentTo("Apple", "Cherry");
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


borland:
This is technically cool as an investigation, but I have reservations around cancellation.

In your example code, the obvious way to implement cancellation would be with something like

```csharp
var queryable = (INevermoreQueryable<Customer>)t.Queryable<Customer>();
await foreach (var customer in queryable)
{
    cancellationToken.ThrowIfCancellationRequested();
}
```

However, in the case of

```csharp
var queryable = (INevermoreQueryable<Customer>)t.Queryable<Customer>()
    .Where(/* very expensive query to evaluate before the first row is returned */)
```

We have no way of cancelling the query without waiting for the first expensive part to complete. For something like an ordered query, SQL Server potentially has to do *all* the work before it's able to return the first row, so this could be quite bad.

While this may not matter too much on the SQL server end (If you cancel a method on the underlying `System.Data.Thing` does that flow through to SQL server in a way that could cancel it partway through execution planning, etc?), on the C# end we still have to wait for all of that to occur.

If this PR merges, then that lets people start using `await foreach` in a way that could fall into this trap. I worry that we may be leaving a rake to be stepped on.

Can we take it in a different way, e.g. something like this should allow for cancel-before-first-row in a way that doesn't leave rakes behind?

```csharp
var queryable = (INevermoreQueryable<Customer>)t.Queryable<Customer>()
    .Where(/* very expensive query to evaluate before the first row is returned */)

// [EnumeratorCancellation] should take care of cancellation within the loop
await foreach (var customer in queryable.AsAsyncEnumerable(cancellationToken))
{
    cancellationToken.ThrowIfCancellationRequested();
}
```

rosslovas:
> Can we take it in a different way, e.g. something like this should allow for cancel-before-first-row in a way that doesn't leave rakes behind?

This is exactly how it's intended to work on the consumer side :slightly_smiling_face: On the Nevermore side, this is just intended to be the _bare minimum_ to make this possible. The consumer can build what you described on top like so:

```csharp
public static async IAsyncEnumerable<T> AsAsyncEnumerable<T>(
    this IQueryable<T> queryable,
    [EnumeratorCancellation] CancellationToken cancellationToken = default)
{
    if (queryable is not INevermoreQueryable<T> awaitableQueryable)
    {
        throw new NotSupportedException("Queryable is not awaitable");
    }

    var enumerator = awaitableQueryable.GetAsyncEnumerator(cancellationToken);
    while (await enumerator.MoveNextAsync(cancellationToken))
    {
        yield return enumerator.Current;
    }
}
```

One thing I could do to better enforce this is simply disallow `await foreach` as is by removing the `= default` and effectively force a consumer to think about cancellation and do something like the above, which is actually what I did at first. The only reason I changed that is because `CancellationToken cancellationToken = default` is unfortunately the norm everywhere in Nevermore, so I felt it was best not to break that trend 😞 

I'd be happy to go against the grain though and remove the `= default` if you think it's necessary to make this change palatable.

borland (approved):
Approve after discussion explaining cancellation and how this fits in with future work in Octopus.Server

PR #252, opened 2023-05-09
Log transaction duration

veochen-octopus:
Log how long each transaction was open for. This follows the same pattern for logging the query duration. The server can then hook into the `ITransactionLogger` to provide telemetry data.

1 of 2 PRs for [sc-47869]. The server PR to be merged after this is https://github.com/OctopusDeploy/OctopusDeploy/pull/17998.

### DIFF
--- a/source/Nevermore.IntegrationTests/RelationalTransaction/TransactionTimerFixture.cs
+++ b/source/Nevermore.IntegrationTests/RelationalTransaction/TransactionTimerFixture.cs
@@ -1,0 +1,43 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using Nevermore.Advanced;
+using Nevermore.Diagnostics;
+using Nevermore.IntegrationTests.SetUp;
+using NUnit.Framework;
+
+namespace Nevermore.IntegrationTests.RelationalTransaction;
+
+public class TransactionTimerFixture : FixtureWithRelationalStore
+{
+    class MockTransactionLogger : ITransactionLogger
+    {
+        public List<(long duration, string transactionName)> Entries { get; } = new();
+        
+        public void Write(long duration, string transactionName)
+        {
+            Entries.Add((duration, transactionName));
+        }
+    }
+    
+    [Test]
+    public void Duration()
+    {
+        var mockTransactionLogger = new MockTransactionLogger();
+        Store.Configuration.TransactionLogger = mockTransactionLogger;
+
+        using (var t = Store.BeginTransaction(name: "timed transaction") as ReadTransaction)
+        {
+            if (t is null)
+            {
+                Assert.Fail($"Transaction is not {nameof(ReadTransaction)}");
+            }
+        
+            t.ExecuteScalar<object>("WAITFOR DELAY '00:00:10'");
+        }
+
+        mockTransactionLogger.Entries.Should().ContainSingle();
+        mockTransactionLogger.Entries.Single().duration.Should().BeGreaterThan(10_000);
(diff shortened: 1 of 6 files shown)


source/Nevermore.IntegrationTests/RelationalTransaction/TransactionTimerFixture.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    class MockTransactionLogger : ITransactionLogger
+    {
+        public List<(long duration, string transactionName)> Entries { get; } = new();
+        
+        public void Write(long duration, string transactionName)
+        {
+            Entries.Add((duration, transactionName));
+        }
+    }
```
veochen-octopus, line 21: No `Moq` or `Substitute` available so made a simple mock logger to verify the method calls.

source/Nevermore/Diagnostics/ITransactionLogger.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Nevermore.Diagnostics
+{
+    public interface ITransactionLogger
+    {
+        void Write(long duration, string transactionName);
```
veochen-octopus, line 5: While this could be another method in `IQueryLogger` which would make things a lot simpler in the short term, a separate interface feels cleaner and easier to work with in the future.

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -66,19 +66,23 @@ public ReadTransaction(IRelationalStore store, RelationalTransactionRegistry reg
...
             connection = new SqlConnection(registry.ConnectionString);
             connection.OpenWithRetry();
+            TransactionTimer = new TimedSection(ms => configuration.TransactionLogger.Write(ms, name));
         }
 
         public async Task OpenAsync()
         {
             connection = new SqlConnection(registry.ConnectionString);
             await connection.OpenWithRetryAsync().ConfigureAwait(false);
+            TransactionTimer = new TimedSection(ms => configuration.TransactionLogger.Write(ms, name));
```
droyad, line 85: I was thinking of just using Stopwatch as `TimedSection` was designed to use with `using` and a single method, however it looks like that it's used in this way for query logging, so better stick to that pattern.

droyad:
We should give Productive Engineering a heads up before merging.

source/Nevermore/Diagnostics/DefaultTransactionLogger.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public DefaultTransactionLogger(long infoThreshold = 10_000)
+        {
+            this.infoThreshold = infoThreshold;
+        }
+        
+        public void Write(long duration, string transactionName)
+        {
+            var level = duration >= infoThreshold ? LogLevel.Info : LogLevel.Debug;
+            Log.Log(level, () => $"Transaction '{transactionName}' was open for {duration}ms");
```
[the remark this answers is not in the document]
veochen-octopus: Yeah saw this in `DefaultQueryLogger` which I followed but thought it was weird. I've now made `DefaultTransactionLogger` do nothing but kept the class so that `RelationalStoreConfiguration` constructor has a placeholder to use.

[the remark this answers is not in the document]
droyad: Yeh empty class is 👍 . The DefaultQueryLogger replicated the the behaviour that was there before it was possible to override it.

YuKitsune (approved):
Looks good from PETs perspective, nice one! :shipit:

PR #253, opened 2023-05-09
Enable `SequentialAccess` for queries translated by the `QueryTranslator`

YuKitsune:
# Summary

When reading documents where the JSON is large, 4MB takes about 22 seconds. The time increases quickly from there to hours for 40MB.

It does seem that lots of node CPU is used as well.

This is due to https://github.com/dotnet/SqlClient/issues/593#issuecomment-642831448.

See OctopusDeploy/Issues#8076

[sc-48263]

# Results

When translating a Linq expression to a `PreparedCommand`, the resulting command will now have the `SequentialAccess` and `SingleResult` behaviour flags enabled.

This is also consistent with other usages of the `PreparedCommand` constructor, as well as [Microsoft's own advice for streaming large BLOBs](https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sqlclient-streaming-support).

While I'm in the area, I figured I'd add the `SingleRow` flag to some `ExecuteScalar` methods. 

See OctopusDeploy/OctopusDeploy#17993 for performance results.

### DIFF
--- a/source/Nevermore/Advanced/Queryable/QueryTranslator.cs
+++ b/source/Nevermore/Advanced/Queryable/QueryTranslator.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Data;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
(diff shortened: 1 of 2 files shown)


paulegradie:
@mattocto @droyad

source/Nevermore/Advanced/QueryBuilder.cs
```diff
@@ -552,11 +552,9 @@ async Task<(List<TRecord>, int)> ToListWithCountAsyncCte(int skip, int take, Can
...
             var stream = readQueryExecutor.StreamAsync(selectSource.GenerateSql(), parmeterValues, map =>
             {
-                if (total == 0)
-                {
-                    total = map.Read(reader => int.Parse(reader[countColumnName].ToString()));
-                }
-                return map.Map<TRecord>(string.Empty);
+                var result = map.Map<TRecord>(string.Empty);
+                total = map.Read(reader => int.Parse(reader[countColumnName].ToString()));
+                return result;
```
YuKitsune, line 557: This is necessary because the `SequentialAccess` flag requires columns to be read in order. Reading the count column first breaks this rule.

This is a breaking change, and I'm inclined to remove it and address it at a later date.

adam-mccoy (approved):
LGTM 👍

PR #255, opened 2023-06-09
Allow inherited types in type check in DataModificationQueryBuilder

dylanlerch:
Relaxes the type check in `PrepareDelete` and `GetDocumentParameters` to allow inherited types of the DocumentMap id column key type.

## Details
The key type check on the [`PrepareLoad` in `ReadTransaction`](https://github.com/OctopusDeploy/Nevermore/blob/b22d050365396e975c45f024dbb28de18bbb95b9/source/Nevermore/Advanced/ReadTransaction.cs#L676-L677) allows for inherited types to be passed in because we check that the mapping id column type matches `TKey` rather than the specific runtime type.

In [`PrepareDelete` in `DataModificationQueryBuilder`](https://github.com/OctopusDeploy/Nevermore/blob/master/source/Nevermore/Util/DataModificationQueryBuilder.cs#L127-L129) called by `WriteTransaction`, we do an exact comparison against the runtime type of the identifier. If it is not an exact type match, this will throw. So while we can support inherited id types for reads, they'll break on delete. 

This change modifies the type check from a exact comparison to an `IsAssignableFrom` comparison. We're still checking the runtime type, but the behaviour is more like what we get with `PrepareLoad`. I've gone for this approach as changing the interface to align with `ReadTransaction` and pass through `TKey` would require an API change. This seems like the least disruptive option as it just makes the type check more permissive and aligns with other behaviour in Nevermore.

### DIFF
--- a/source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.DeleteByIdWithInheritedTypedKey.approved.txt
+++ b/source/Nevermore.Tests/Util/DataModificationQueryBuilderFixture.DeleteByIdWithInheritedTypedKey.approved.txt
@@ -1,0 +1,3 @@
+DELETE FROM [dbo].[TestDocumentWithTypedKey] WITH (ROWLOCK) WHERE [Id] = @Id
+
+@Id=Nevermore.Tests.Util.DataModificationQueryBuilderFixture+InheritedTypedKey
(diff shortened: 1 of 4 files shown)


adam-mccoy (approved):
LGTM

borland (approved):
LGTM, the change is simple and better all round as it makes it consistent between read and modify. I appreciate the good PR description, thanks!

PR #259, opened 2023-07-10
Add a Roslyn Analyzer for unsupported IQueryable methods

YuKitsune:
Adds a Roslyn Analyzer that raises compiler errors when an unsupported `IQueryable` method is invoked. This analyzer only targets calls to Nevermore types, i.e: `IReadQueryExecutor.Queryable<T>().Single()` and doesn't affect other calls to `IQueryable`, i.e: `myList.AsQueryable().Single()`.

Also makes the visitors public so that they can be accessed in Octopus Server and re-purposed.

[sc-53501]


borland (approved):
LGTM.
The tests not following the standard patterns we have in Octopus.Server.Analyzers with the `{|#0:` is a bit weird, but I can see you're following precedent within this repo; we can worry about that later.

PR #260, opened 2023-07-10
ReadTransaction will now re-open a SqlConnection if the underlying one is not open.

borland:
# Background

The Octopus Cloud Team noticed a number of `InvalidOperationException` being thrown when they reached the request limit for an elastic pool. This originates from the BeginTransaction operation, stating that the operation is invalid because the connection is closed.

The theory of why this happens is:

1. We attempt to begin a transaction
2. The request fails due to hitting the limit (a transient error), tearing down the connection
3. We retry beginning the transaction without first re-establishing the connection

Similar to how retrying commands work, retrying a transaction should involve ensuring there is a valid connection, and re-establishing one if not.

## Note regarding testing:

- I was not able to manually test the error-handling path; There's no way to cause a real SQL connection to purposefully throw a transient error on BeginTransaction.

- The Nevermore integration tests talk to an actual SQL server database with a real `SqlConnection`, so as per the manual test scenario, we can't trigger an error in order to test the error handling.

Discussed using something like [Toxiproxy](https://github.com/Shopify/toxiproxy) however to get this to work it seemed like I'd end up needing to write a Toxiproxy (or other proxy-thing) plugin to parse the TDS binary data stream and inject an error at an appropriate place. While this was doable, it felt like too much work in not quite the right direction.

Instead, I decided to pivot and refactor nevermore to allow a mocked-out `SqlConnection`. With this in place I wrote some unit tests which exercise the retry behavior. It remains to be seen whether this will catch the exact specific error the cloud team observed, but if we've missed it then it should be trivial to update the unit tests to reflect reality and then fix the code.

**Note** I rebuilt my local Octopus server against this refactored Nevermore and ran a variety of integration tests, thus putting a lot of data through the Nevermore stack. Everything worked well.


source/Nevermore.Analyzers/Descriptors.cs
```diff
@@ -1,8 +1,11 @@
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.CodeAnalysis;
 // ReSharper disable InconsistentNaming
 
 namespace Nevermore.Analyzers
 {
+    // https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.Analyzers/ReleaseTrackingAnalyzers.Help.md
+    [SuppressMessage("Roslyn", "RS2008:AnalyzerReleaseTracking", Justification = "We choose not to follow Microsoft's Analyzer Release tracking scheme")]
```
borland, line 8: Cleaning up unrelated but trivial lint-warnings while I was in the codebase

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -665,6 +666,38 @@ protected async Task<TResult[]> ReadResultsAsync<TResult>(PreparedCommand prepar
             using var command = CreateCommand(preparedCommand);
             return await command.ReadResultsAsync(mapper, cancellationToken).ConfigureAwait(false);
         }
+        
+        async Task<DbTransaction> BeginTransactionWithRetryAsync(IsolationLevel isolationLevel, string sqlServerTransactionName, RetryPolicy retryPolicy)
```
borland, line 670: These moved out of the now-deleted `TransactionExtensions` in order to allow reassignment of `connection`.
The extensions weren't used anywhere else and arguably it's better to have them as private helpers here anyway

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -810,6 +823,16 @@ void ITransactionDiagnostic.WriteCurrentTransactions(StringBuilder output)
             registry.WriteCurrentTransactions(output);
         }
 
+        DbTransaction BeginTransaction(IsolationLevel isolationLevel, string sqlServerTransactionName)
+        {
+            if (connection is SqlConnection sqlConnection)
+            {
+                return sqlConnection.BeginTransaction(isolationLevel, sqlServerTransactionName);
```
borland, line 830: @adam-mccoy this here (`DbConnection` not having a named `BeginTransaction` overload) was why I made the ISqlConnection interface rather than passing DbConnection around the place. I don't disagree with your change here (it felt like a bit of a coin-toss when I made the interface) but it'd be interesting to get your thoughts about why

source/Nevermore.IntegrationTests/Nevermore.IntegrationTests.csproj
```diff
@@ -13,6 +13,11 @@
     <IsPackable>false</IsPackable>
     <WarningsAsErrors />
     <TargetFramework>net6.0</TargetFramework>
+    <!-- 
+    @team-productive-engineering NV0002 is designed to catch unsupported Queryable calls in Octopus Server,
+    but isn't quite right at the moment when pointed at nevermore itself. Suppress for now 
+    -->
+    <NoWarn>NV0002</NoWarn>
```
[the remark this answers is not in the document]
borland: @adam-mccoy I think @YuKitsune may have fixed up the analyzer so this is no longer necessary. Can we give it a spin without?

adam-mccoy (approved):

PR #261, opened 2023-07-12
Ignore non-IQueryable methods

YuKitsune:
Updates the `IQueryable` analyser to ignore non `IQueryable` methods, e.g: `.ToList()`.


source/Nevermore.Analyzers.Tests/NevermoreQueryableAnalyzerFixture.cs
```diff
@@ -4,6 +4,18 @@ namespace Nevermore.Analyzers.Tests;
 
 public class NevermoreQueryableAnalyzerFixture : NevermoreFixture
 {
+	[Test]
```
desruc, line 7: supernit: this test tabbed out further than the others.

desruc (approved):
Changes look good to me 🙂 Nice one.

PR #262, opened 2023-07-25
Escape `LIKE` comparison so that `[` and `%` can be used in queries

emilol:
When querying a value with `.Contains`, the resulting SQL is a `WHERE foo LIKE %{value}%` clause. When `value` contains special characters (`[` or `%`) the results can be unpredictable.

### DIFF
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -669,6 +669,32 @@ namespace Nevermore.IntegrationTests
 
             customers.Select(c => c.FirstName).Should().BeEquivalentTo("Alice", "Bob", "Charlie");
         }
+        
+        [Test]
+        public void WhereStringContains()
+        {
+            using var t = Store.BeginTransaction();
+
+            var testCustomers = new[]
+            {
+                new Customer { FirstName = "Alice", LastName = "Apple", Nickname = "Bear" },
+                new Customer { FirstName = "Bob", LastName = "Banana", Nickname = "Beets" },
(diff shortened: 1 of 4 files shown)


YuKitsune (approved):
Nice find!

source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
```diff
@@ -669,6 +669,32 @@ public void WhereNotContainsEmpty()
 
             customers.Select(c => c.FirstName).Should().BeEquivalentTo("Alice", "Bob", "Charlie");
         }
+        
+        [Test]
+        public void WhereStringContains()
```
emilol, line 674: We had `WhereNotStringContains` but no`WhereStringContains` so I added it while I was here

source/Nevermore/LikeComparisonExtensions.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Nevermore
+{
+    public static class LikeComparisonExtensions {
+        public static string EscapeForLikeComparison(this string value)
+        {
+            return value.Replace("[", "[[]").Replace("%", "[%]");
```
adam-mccoy, line 6: Underscore is also a wildcard character, per [the docs](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql?view=sql-server-ver16#use-wildcard-characters-as-literals). Perhaps we should handle that too?

adam-mccoy (approved):
Good find 👍
Left one suggestion that's been missing forever.

source/Nevermore/LikeComparisonExtensions.cs
YuKitsune: Good catch, I've added a case for this 👍

PR #264, opened 2023-07-27
Correctly identify DateTimeOffset JSON properties as scalar

APErebus:
When performing a `.Where` query against a property of type `DateTimeOffset`, it was incorrectly being identified as a non-scalar type, meaning nevermore tried to use `JSON_QUERY`, as opposed to `JSON_VALUE`.

This is because the TypeCode returned is `Object`.

This PR just updates the `IsScalar` to check certain predefined scalar types before falling back on the `GetTypeCode` check


adam-mccoy (approved):
LGTM. Nice pickup 👍

PR #265, opened 2023-08-04
Async and cancellation improvements

rosslovas:
Cancellation tokens weren't being passed to `OpenAsync`. This means that failure to acquire a connection (e.g. timing out trying to obtain a connection from the SQL connection pool) cannot be cancelled. This PR resolves that.

### DIFF
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -684,17 +684,19 @@ namespace Nevermore.Advanced
         {
             if (connection == null) throw new InvalidOperationException("Must create a DbConnection before attempting to begin a transaction");
 
-            return await retryPolicy.LoggingRetries("Beginning Database Transaction").ExecuteActionAsync(async () =>
-            {
-                // A connection can exist, but be in a broken state.
-                // E.g. a valid connection is returned to the pool, we then acquire it, but on the SQL server end it's been killed perhaps due to Azure SQL resource limits.
-                // We re-open the same connection, following the logic in `DbCommandExtensions.EnsureValidConnection`
-                if (connection.State != ConnectionState.Open) await connection.OpenAsync(cancellationToken).ConfigureAwait(false);
+            return await retryPolicy.LoggingRetries("Beginning Database Transaction").ExecuteActionAsync(
+                async ct =>
+                {
+                    // A connection can exist, but be in a broken state.
+                    // E.g. a valid connection is returned to the pool, we then acquire it, but on the SQL server end it's been killed perhaps due to Azure SQL resource limits.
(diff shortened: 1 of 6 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -90,15 +90,15 @@ public void Open()
 
             connection = connectionFactory(registry.ConnectionString);
             connection.OpenWithRetry();
-            
+
             TransactionTimer = new TimedSection(ms => configuration.TransactionLogger.Write(ms, name));
         }
 
-        public async Task OpenAsync()
+        public async Task OpenAsync(CancellationToken cancellationToken = default)
```
borland, line 97: This is an ABI break and could cause a crash if an old DLL indirectly gets handed a new version of Nevermore (e.g. multiple projects in the same solution reference different versions).
I don't expect we care about such things for Nevermore, just felt I should raise it just in in case.

source/Nevermore/Transient/DbCommandExtensions.cs
```diff
@@ -31,7 +31,7 @@ public static Task<int> ExecuteNonQueryWithRetryAsync(this DbCommand command, Re
         {
             GuardConnectionIsNotNull(command);
             var effectiveCommandRetryPolicy = (commandRetryPolicy ?? RetryPolicy.NoRetry).LoggingRetries(operationName);
-            return effectiveCommandRetryPolicy.ExecuteAction(async () =>
+            return effectiveCommandRetryPolicy.ExecuteActionAsync(async () =>
```
borland, line 34: Oooh nice catch. Async lambda's are dangerous.

I expected to see the cancellationToken passed through to `ExecuteActionAsync` though. Is that a concern or is it fine because the stuff inside the lambda shares the token anyway?

borland (approved):

source/Nevermore/Transient/DbCommandExtensions.cs
rosslovas: >I expected to see the cancellationToken passed through to `ExecuteActionAsync` though. Is that a concern or is it fine because the stuff inside the lambda shares the token anyway?

Yeah me too. It should be fine because `ExecuteActionAsync` doesn't actually do anything async, no delays or anything, it just immediately retries. I suppose there is potential benefit to taking it and adding in a `ThrowIfCancellationRequested`, I'm still considering whether I want to add that now or leave it for the time being 🤔

rosslovas: I decided to make this change in the end, though I've left overloads that don't take a cancellation token around so that this is not a breaking change as this stuff is public.

borland (approved):

PR #267, opened 2023-08-11
Add AllocateIdAsync and AsyncPrimaryKeyHandler

rosslovas:
This adds async & cancellable overloads to `IWriteQueryExecutor` for `AllocateId`. To accomplish this, it also adds `IAsyncPrimaryKeyHandler` and `AsyncPrimaryKeyHandler` for consumers to (optionally) implement/override. If a primary key handler doesn't do so (and this is entirely valid e.g. a primary key handler like `GuidPrimaryKeyHandler` has no reason to be async) then the synchronous methods will be used instead as they always had previously.

This PR does not attempt to make ID allocation async all throughout Nevermore as that's a fairly big change that can be done independently of this one, so is best left to its own PR.

This is not a breaking change if you implement/override `IPrimaryKeyHandler` / `PrimaryKeyHandler`, as these are designed for consumers to implement/override. It's entirely optional to implement/override `IAsyncPrimaryKeyHandler` / `AsyncPrimaryKeyHandler`; if you don't, the synchronous methods will be used instead.

This _is_ a breaking change if a consumer implements `IKeyAllocator` or `IWriteQueryExecutor` as there are a couple async overloads to implement, but I think it's unlikely anyone is doing so, and it's a minor & easy to resolve change either way.


source/Nevermore/Mapping/KeyAllocator.cs
```diff
@@ -28,15 +31,21 @@ public int NextId(string tableName)
...
             readonly string collectionName;
             readonly int blockSize;
-            readonly object sync = new object();
-            volatile int blockStart;
-            volatile int blockNext;
-            volatile int blockFinish;
+            readonly SemaphoreSlim sync = new(1, 1);
+            int blockStart;
+            int blockNext;
+            int blockFinish;
```
rosslovas, line 48: The `volatile` keyword wasn't serving any purpose here. These were always protected by the `sync` mutex.

source/Nevermore/Mapping/KeyAllocator.cs
```diff
@@ -28,15 +31,21 @@ public int NextId(string tableName)
             return allocation.Next();
         }
 
+        public async Task<int> NextIdAsync(string tableName, CancellationToken cancellationToken)
```
borland, line 34: Given that this holds blocks, 19 out of 20 times it's going to return synchronously. As such, might be nice to return ValueTask (and plumb that back up the stack through AllocateNextAsync) to communicate that

rosslovas: Yeah that occurred to me as I wrote this, I opted for `Task` anyway due to the "consume at most once" restriction on `ValueTask` feeling a bit dangerous but you're right it does seem like a good fit. I'm still not sure how to evaluate that.

One point in favour of choosing `ValueTask` here is that it's relatively safe to later on go `ValueTask` -> `Task` (might break at compile time on `AsTask()`, but it won't break anything at runtime at least) but not safe to go the other way. 🤔

rosslovas: I decided to go with `ValueTask` in the end 👍

borland (approved):
LGTM!

PR #268, opened 2023-09-28
Log transactions every minute (when there are a high number active)

desruc:
The `RelationalTransactionRegistry` is responsible for logging warnings when a high number of active transactions is detected.

If the number of transactions reaches 90% of the maximum pool size or exceeds the maximum size, a comprehensive list of transactions is logged. This check happens when adding another transaction to the registry.

Currently, there's a `highNumberAlreadyLoggedAtError` flag that determines whether this list should be logged. The flag is set to true after the first log, preventing subsequent logs.

This PR replaces the flag with a `DateTime` and refactors the logic to log the list of transactions every minute if the high transaction count condition is met.


rosslovas (approved):
Solid improvement, should help a lot!

PR #269, opened 2023-10-10
Allow using a custom key allocator by setting the `KeyAllocatorFactory` configuration

tekerson:
This allows the configuration of a custom `IKeyAllocator` implementation which will be used to provide the numerical suffix when generating Ids. We intend to use this to synchronise key allocation with documents that are outside of the control of Nevermore.


emilol (approved):
Nice and simples. LGTM

PR #271, opened 2023-10-26
Support IQueryable Projection

adam-mccoy:
# Background

Add basic support for projection in IQueryable queries.

## Project a single column

```cs
transaction.Queryable<Document>().Select(d => d.Id);
```

## Project a set of properties as an anonymous type

```cs
transaction.Queryable<Document>().Select(d => new { d.Id, d.Name });
```

### DIFF
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1527,5 +1527,29 @@ namespace Nevermore.IntegrationTests
                 new { FirstName = "Charlie", IsEmployee = true }
             });
         }
+
+        [Test]
+        public async Task ProjectValueType()
+        {
+            using var t = Store.BeginTransaction();
+
+            var testCustomers = new[]
+            {
+                new Customer { FirstName = "Alice", LastName = "Apple", IsEmployee = true },
+                new Customer { FirstName = "Bob", LastName = "Banana", IsEmployee = false },
+                new Customer { FirstName = "Charlie", LastName = "Cherry", IsEmployee = true }
+            };
+
+            foreach (var c in testCustomers)
+            {
+                t.Insert(c);
+            }
+
+            await t.CommitAsync();
+
+            var customers = await t.Queryable<Customer>().Select(c => c.IsEmployee).ToListAsync();
+
+            customers.Should().BeEquivalentTo(new[] { true, false, true });
+        }
     }
 }
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1551,5 +1551,42 @@ namespace Nevermore.IntegrationTests
 
             customers.Should().BeEquivalentTo(new[] { true, false, true });
         }
+
+        [Test]
+        public async Task ProjectTypeWithParameterizedConstructor()
+        {
+            using var t = Store.BeginTransaction();
+
+            var testCustomers = new[]
+            {
+                new Customer { FirstName = "Alice", LastName = "Apple", IsEmployee = true },
+                new Customer { FirstName = "Bob", LastName = "Banana", IsEmployee = false },
+                new Customer { FirstName = "Charlie", LastName = "Cherry", IsEmployee = true }
+            };
+
+            await t.InsertManyAsync(testCustomers);
+            await t.CommitAsync();
+
+            var customers = await t.Queryable<Customer>().Select(c => new CustomerProjection(c.FirstName, c.IsEmployee)).ToListAsync();
+
+            customers.Should().BeEquivalentTo(new[]
+            {
+                new CustomerProjection("Alice", true),
+                new CustomerProjection("Bob", false),
+                new CustomerProjection("Charlie", true)
+            });
+        }
+
+        class CustomerProjection
+        {
+            public CustomerProjection(string firstName, bool isEmployee)
+            {
+                FirstName = firstName;
+                IsEmployee = isEmployee;
+            }
+
+            public string FirstName { get; }
+            public bool IsEmployee { get; }
+        }
     }
 }
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1521,10 +1521,11 @@ namespace Nevermore.IntegrationTests
 
             var customers = t.Queryable<Customer>().Select(c => new { c.FirstName, c.IsEmployee });
 
-            customers.Should().BeEquivalentTo(
+            customers.Should().BeEquivalentTo(new[] {
                 new { FirstName = "Alice", IsEmployee = true },
                 new { FirstName = "Bob", IsEmployee = false },
-                new { FirstName = "Charlie", IsEmployee = true });
+                new { FirstName = "Charlie", IsEmployee = true }
+            });
         }
     }
 }
--- a/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
+++ b/source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
@@ -1475,5 +1475,56 @@ namespace Nevermore.IntegrationTests
 
             customers.Select(c => c.LastName).Should().BeEquivalentTo("Apple", "Cherry");
         }
+
+        [Test]
+        public async Task ProjectColumnField()
+        {
+            using var t = Store.BeginTransaction();
+
+            var testCustomers = new[]
+            {
+                new Customer { FirstName = "Alice", LastName = "Apple", Roles = { "Admin" }},
+                new Customer { FirstName = "Bob", LastName = "Banana", Roles = { "Boss" } },
+                new Customer { FirstName = "Charlie", LastName = "Cherry", Roles = { "Editor", "Bum" } }
+            };
+
+            foreach (var c in testCustomers)
+            {
+                t.Insert(c);
+            }
+
+            await t.CommitAsync();
+
+            var customers = t.Queryable<Customer>().Select(c => c.FirstName);
+
+            customers.Should().BeEquivalentTo("Alice", "Bob", "Charlie");
+        }
+
+        [Test]
+        public async Task ProjectJsonField()
+        {
+            using var t = Store.BeginTransaction();
+
+            var testCustomers = new[]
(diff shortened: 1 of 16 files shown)


source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
```diff
@@ -1475,5 +1475,118 @@ public async Task AsyncEnumeration()
...
+
+            await t.CommitAsync();
+
+            var customers = await t.Queryable<Customer>().Select(c => c.IsEmployee).ToListAsync();
+
+            customers.Should().BeEquivalentTo(new[] { true, false, true });
+        }
+
+        [Test]
+        public async Task ProjectTypeWithParameterizedConstructor()
```
borland, line 1556: I'm pretty sure that `ProjectTypeWithParameterizedConstructor` ends up being the same, but could you add a `ProjectIntoRecordType` test just to clear that pathway? If you're feeling keen a `ProjectIntoRecordStructType` too :smile:

source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
```diff
@@ -1475,5 +1475,118 @@ public async Task AsyncEnumeration()
...
+                new Customer { FirstName = "Bob", LastName = "Banana", IsEmployee = false },
+                new Customer { FirstName = "Charlie", LastName = "Cherry", IsEmployee = true }
+            };
+
+            await t.InsertManyAsync(testCustomers);
+            await t.CommitAsync();
+
+            var customers = await t.Queryable<Customer>().Select(c => new CustomerProjection(c.FirstName, c.IsEmployee)).ToListAsync();
+
+            customers.Should().BeEquivalentTo(new[]
```
borland, line 1572: BeEquivalentTo doesn't care about row ordering. Is that significant here? (you can add it by setting options on the BeEquivalentTo)

source/Nevermore/Advanced/Queryable/AsyncEnumerableAdapter.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Nevermore.Advanced.Queryable
+{
+    internal static class AsyncEnumerableAdapter
+    {
+        public static async Task<IEnumerable> ConvertToEnumerable(object asyncEnumerable, Type sequenceType, CancellationToken cancellationToken)
+        {
+            var asyncEnumerableType = typeof(IAsyncEnumerable<>).MakeGenericType(sequenceType);
+            var asyncEnumeratorType = typeof(IAsyncEnumerator<>).MakeGenericType(sequenceType);
+
+            var getAsyncEnumeratorMethod = asyncEnumerableType.GetRuntimeMethod("GetAsyncEnumerator", new[] { typeof(CancellationToken) });
```
borland, line 17: 🤔  I hadn't seen `GetRuntimeMethod` used before and it doesn't appear anywhere in the Octopus.Server codebase.

Decompiler shows it's just a wrapper which calls `type.GetMethod`, but the Microsoft Docs are useless and don't offer any explanation as to why there are two methods which do the same thing.

It does appear to be used in other places in Nevermore so keep-calm-carry on I guess?

source/Nevermore/Advanced/Queryable/AsyncEnumerableAdapter.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nevermore.Advanced.Queryable
+{
+    internal static class AsyncEnumerableAdapter
+    {
+        public static async Task<IEnumerable> ConvertToEnumerable(object asyncEnumerable, Type sequenceType, CancellationToken cancellationToken)
+        {
+            var asyncEnumerableType = typeof(IAsyncEnumerable<>).MakeGenericType(sequenceType);
```
borland, line 14: A tip for reducing the amount of generic hoop-jumping:

Create an actual generic method which does what you want, then instead of doing lots of `MakeGenericType`, do a `MakeGenericMethod` and call it via reflection.

This lets you get rid of CreateList too

e.g.

```csharp
static async Task<IEnumerable> ConvertToEnumerableHelper<T>(IAsyncEnumerable<T> enumerable, CancellationToken ct) {
    var result = new List<T>();
    await foreach(var item in enumerable.WithCancellation(ct)) {
       result.add(item);
    }
    return result;
}

public static Task<IEnumerable> ConvertToEnumerable(object asyncEnumerable, Type sequenceType, CancellationToken cancellationToken) {
   var method = typeof(AsyncEnumerableAdapter).GetMethod("ConvertToEnumerableHelper").MakeGenericMethod(sequenceType);
   
   return (Task<IEnumerable>)method.Invoke(asyncEnumerable);
}
```

source/Nevermore/Advanced/Queryable/QueryProvider.cs
```diff
@@ -121,31 +103,37 @@ public IAsyncEnumerable<TResult> StreamAsync<TResult>(Expression expression, Can
...
-#pragma warning disable CA2007
-            // CA2007 doesn't understand ConfiguredCancelableAsyncEnumerable and incorrectly thinks we need another ConfigureAwait(false) here
-            await using var enumerator = enumerable.ConfigureAwait(false).WithCancellation(cancellationToken).GetAsyncEnumerator();
-#pragma warning restore CA2007
+            return GenericExecuteScalarMethod.MakeGenericMethod(resultType)
+                .Invoke(queryExecutor, new object[] { command });
+        }
 
-            if (await enumerator.MoveNextAsync())
+        static object FirstOrDefault(IEnumerable stream)
```
borland, line 126: should be `object?` for null-safety

source/Nevermore/Advanced/Queryable/QueryTranslator.cs
```diff
@@ -183,11 +183,41 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
...
+                    Visit(node.Arguments[0]);
+                    AddProjection(node.Arguments[1]);
+                    return node;
+                }
                 default:
                     throw new NotSupportedException();
             }
         }
 
+        void AddProjection(Expression expression)
```
borland, line 197: Disclaimer: A solid review of this requires a pretty deep understanding of what's going on. I feel like I'm about 75% there, but not all the way, so I'm mostly pattern-matching on code structure and sanity checking things; I'm highly unlikely to spot any actual bugs or correctness problems and I'm leaning on your unit tests to cover those.

source/Nevermore/Advanced/ReaderStrategies/ArbitraryClasses/ArbitraryClassReaderStrategy.cs
```diff
@@ -81,18 +78,31 @@ public bool CanRead(Type type)
...
-            
-            // Find the parameter-less constructor
+
             var constructors = typeof(TRecord).GetConstructors();
-            var defaultConstructor = constructors.FirstOrDefault(p => p.GetParameters().Length == 0) ?? throw new InvalidOperationException("When reading query results to a class, the class must provide a default constructor with no parameters.");
-            
-            // Create fast setters for all properties on the type
-            var properties = typeof(TRecord).GetProperties(BindingFlags.Public | BindingFlags.Instance | BindingFlags.SetProperty);
-            
+            // Find the parameter-less constructor
```
borland, line 95: `GetConstructors` will only find public constructors. Should we consider specifying BindingFlags to find any protected/private constructors too? (I think if we did, it'd be good to sort them such that public constructors are matched first). Probably on balance though best left alone?

source/Nevermore/Querying/AST/ISelectColumns.cs
```diff
@@ -110,4 +111,53 @@ public SelectRowNumber(Over over, string alias)
...
+        {
+            this.jsonPath = jsonPath;
+            this.elementType = elementType;
+        }
+
+        public bool AggregatesRows => false;
+
+        public string GenerateSql() => $"CONVERT({GetDbType()}, JSON_VALUE([JSON], '{jsonPath}'))";
+
+        string GetDbType()
```
borland, line 144: This function already exists in `Nevermore/Querying/AST/Where.cs` (I'm assuming that's where you got it from).

It seems significant enough to extract into a static helper that we can re-use?

borland (approved):
LGTM: I've left some suggestions that might improve the PR; please consider and merge when you're happy.

source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
adam-mccoy: I think you're right in that records would be the same, but it doesn't hurt to add an extra test to prove it. Structs aren't currently supported at all, so allowing `record struct` would be more difficult. Probably not worth taking on at this stage.

source/Nevermore.IntegrationTests/QueryableIntegrationFixture.cs
adam-mccoy: I don't think it really matters, since what this test cares about is the successful projection. Still, if a more restrictive assertion like `Equal()` can be used it should be.

borland: Actually fair enough to use the unordered check here. SQL server doesn't guarantee row order unless you have an OrderBy clause, and you don't have one here. Adding that in the mix would be a different test.

source/Nevermore/Advanced/Queryable/AsyncEnumerableAdapter.cs
adam-mccoy: Good call. I'll tidy it up

source/Nevermore/Advanced/Queryable/QueryProvider.cs
adam-mccoy: Nevermore doesn't have #nullable turned on currently. Another problem for another day 😄

source/Nevermore/Advanced/ReaderStrategies/ArbitraryClasses/ArbitraryClassReaderStrategy.cs
adam-mccoy: I agree it'd be handy, but status-quo is fine for now.

source/Nevermore/Querying/AST/ISelectColumns.cs
adam-mccoy: I had actually already done this, but just hadn't pushed it 😁

PR #273, opened 2023-11-07
Add support for adopting an existing connection/transaction

rosslovas:
# Background

This allows Nevermore to work with an existing connection/transaction pair rather than always creating and owning its own. This allows multiple Nevermore transactions to be used as part of a single SQL transaction, or sharing a single transaction between multiple ORMs, or simply enabling the ownership of the connection/transaction to remain with the consumer.

This is similar to EF's `UseTransaction`: https://learn.microsoft.com/en-us/ef/core/saving/transactions#share-connection-and-transaction

If an external connection is provided, the consumer is responsible for commit/dispose of the connection/transaction.

# Notes

**This is a breaking change as there are changes to public APIs.** Nevermore's major version will be incremented as part of this change.

Notably, there are new arguments to the `ReadTransaction` and `WriteTransaction` constructors as well as whole new constructor overloads, and the `IRelationalStore` interface now has two new methods
* `CreateReadTransactionFromExistingConnectionAndTransaction`
* `CreateWriteTransactionFromExistingConnectionAndTransaction`


source/Nevermore/Advanced/IConnectionAndTransaction.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Nevermore.Advanced
+{
+    public interface IConnectionAndTransaction : IDisposable
+    {
+        DbConnection Connection { get; }
+        DbTransaction Transaction { get; }
+        void CommitTransaction();
+        Task CommitTransactionAsync(CancellationToken cancellationToken);
+    }
```
rosslovas, line 14: The benefit of this interface is it gives the consumer full control over what happens on commit or dispose e.g. they may want to do some logging or something.

source/Nevermore/IsExternalInit.cs
```diff
@@ -0,0 +1,8 @@
+// Hack to let us use records targeting netstandard2.1
+// ReSharper disable once CheckNamespace
+namespace System.Runtime.CompilerServices
+{
+    static class IsExternalInit
```
rosslovas, line 5: More context on this: https://stackoverflow.com/a/67687186/19295604

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -586,6 +644,8 @@ void AddCommandTrace(string commandText)
                 if (commandTrace.Count <= 200)
                     commandTrace.Add(DateTime.Now.ToString("s") + " " + commandText);
             }
+
+            customCommandTrace?.Invoke(commandText);
```
rosslovas, line 648: When this is provided, we _could_ disable the default command tracing behaviour and leave it up to the consumer only. However, other things that are tied to this still exist like `RelationalTransactionRegistry`, and there's no guarantee the consumer doesn't still want that stuff. So, I think it's safest to just leave it be.

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,167 @@
...
+        var sqlTransaction = sqlConnection.BeginTransaction();
+        using var connectionAndTransaction = new TestConnectionAndTransaction(sqlConnection, sqlTransaction);
+
+        var nonOwnedTransaction = Store.CreateWriteTransactionFromExistingConnectionAndTransaction(
+            connectionAndTransaction, false);
+
+        Assert.Multiple(() =>
+        {
+            Assert.ThrowsAsync<InvalidOperationException>(async () => await nonOwnedTransaction.CommitAsync(cancellationToken));
+            Assert.Throws<InvalidOperationException>(() => nonOwnedTransaction.Commit());
```
borland, line 114: InvalidOperationException can happen for lots of reasons. As above suggest FluentAssertions, then you could do `.WithMessage()` and assert that you got the right one.

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -26,16 +26,29 @@ namespace Nevermore.Advanced
...
+        readonly IRelationalTransactionRegistry registry;
         readonly RetriableOperation operationsToRetry;
         readonly IRelationalStoreConfiguration configuration;
         readonly ITableAliasGenerator tableAliasGenerator = new TableAliasGenerator();
 
         readonly Func<string, DbConnection> connectionFactory;
+        readonly Action<string>? customCommandTrace;
         readonly string name;
 
+        protected abstract record ExistingConnectionAndTransaction(IConnectionAndTransaction Value)
```
borland, line 40: An [ADT](https://en.wikipedia.org/wiki/Algebraic_data_type) where all cases hold the same value feels a bit odd, maybe it's not the best tool for the job here?

The fact that the class can hold both an `ExistingConnectionAndTransaction` _and_ ` DbConnection` at the same time feels a bit weird too. It allows for this to hold two transactions at once which doesn't seem quite right?

Looking at the code, could we make it more idiomatic C# with something like the following:
It might clean things up a little bit too?

- Make a nevermore-internal class NevermoreConnectionAndTransaction : IConnectionAndTransaction to hold the places where Nevermore creates the transaction itself
- ReadTransaction always holds an IConnectionAndTransaction, rather than the nullable "existing" thing.
- This is either the nevermore-internal one or something else (who cares what type it is, doesn't matter)
- there's a `readonly bool ownsTransaction` flag that controls whether we dispose it here or not.

This is probably more in-line with what people would expect to see based on other code like Stream in C# which also has the "sometimes-owned" behaviour

borland (commented):
The big-picture design and implementation looks sound.

I've offered some suggestions on ways you might improve/simplify some things, but I'm happy if you choose to disagree on them.

Will leave approval to the Normalisation sub-group

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,167 @@
...
+        await sqlConnection.OpenAsync(cancellationToken);
+        await using var sqlTransaction = sqlConnection.BeginTransaction();
+
+        var readTransaction = (ReadTransaction) Store.CreateReadTransactionFromExistingConnectionAndTransaction(
+            new TestConnectionAndTransaction(sqlConnection, sqlTransaction),
+            takeOwnershipOfExistingConnectionAndTransaction);
+
+        Assert.Multiple(() =>
+        {
+            Assert.ThrowsAsync<InvalidOperationException>(async () => await readTransaction.OpenAsync(cancellationToken));
```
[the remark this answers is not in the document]
rosslovas: Yeah that's FluentAssertions only, [NUnit waits on it synchronously](https://github.com/nunit/nunit/blob/097a1be976fb6bfff6a2a3a9d8b24dd5010fbcdb/src/NUnitFramework/framework/Assert.Exceptions.Async.cs#L26).

source/Nevermore/Advanced/IConnectionAndTransaction.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Nevermore.Advanced
+{
+    public interface IConnectionAndTransaction : IDisposable
```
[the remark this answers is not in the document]
rosslovas: Nevermore doesn't have a single implementation of `IAsyncDisposable` right now. I decided to leave it out of scope but it could be done as a separate PR for sure.

My implementation of this does implement `IAsyncDisposable`, though it only plays a role if you maintain ownership over it, it doesn't get called if you give Nevermore ownership then dispose the Nevermore transaction.

source/Nevermore/Advanced/ReadTransaction.cs
rosslovas: > An [ADT](https://en.wikipedia.org/wiki/Algebraic_data_type) where all cases hold the same value feels a bit odd, maybe it's not the best tool for the job here?

This was just personal preference, it offers no real advantage; it's functionally identical to something that carries an `IConnectionAndTransaction` along with an `enum Kind { ... }` for example but I just prefer to use the type itself to carry that info if I can. I'd be fine with changing it if there's strong disagreement.

> The fact that the class can hold both an ExistingConnectionAndTransaction and  DbConnection at the same time feels a bit weird too. It allows for this to hold two transactions at once which doesn't seem quite right?
>
> Looking at the code, could we make it more idiomatic C# with something like the following:
It might clean things up a little bit too?

I initially went that route but decided to take this approach purely for the sake of minimising changes. In this PR I was pretty torn between architecting things nicely vs a minimal change approach just to serve our needs. I'd be happy to change it and go that route after all if we decide it's worthwhile to invest in Nevermore like that, I think it's worth a discussion.

source/Nevermore/Advanced/RelationalTransactionRegistry.cs
```diff
@@ -12,14 +12,21 @@
 
 namespace Nevermore.Advanced
 {
-    public class RelationalTransactionRegistry
+    public interface IRelationalTransactionRegistry
+    {
+        void Add(ReadTransaction trn);
```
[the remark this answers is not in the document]
rosslovas: Agree that it could be improved, I chose to keep that out of scope though. FWIW the reason I added the interface is to provide a no-op implementation so we can implement a better alternative completely outside of Nevermore.

This class is a bit odd already, it logs a message with `The below information may help the Octopus team diagnose the problem` which suggests it shouldn't be in Nevermore to begin with.

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            Connection = connection;
+            Transaction = transaction;
+        }
+
+        public DbConnection Connection { get; }
+
+        public DbTransaction Transaction { get; }
+
+        public int CommitTransactionCalled { get; private set; }
```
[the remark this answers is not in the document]
rosslovas: The reason I'm asserting on calls to these methods is specifically because the implementation of `IConnectionAndTransaction` is consumer-provided; it comes from outside Nevermore, so it could be anything. Given that, I care about how many times they're called a lot more than I care what they actually do, and so not asserting on outcomes was a deliberate choice.

I made them "work" by calling `Commit`/`Dispose` only for demonstration & convenience purposes honestly, these tests would work equally well if they just incremented those counters and didn't actually do any work, leaving the actual dispose up to the tests themselves.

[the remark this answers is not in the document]
borland: Makes sense, thanks.

Maybe consider the List<string> approach rather than counters? But not a big deal either way.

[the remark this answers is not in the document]
rosslovas: I think the list approach is reasonable and has worthwhile benefits, I'll give it a go 🙂

source/Nevermore/Advanced/ReadTransaction.cs
borland: > it's functionally identical to something that carries an IConnectionAndTransaction along with an enum Kind { ... }

Yep I am well acquainted with this pattern. 

The main thing that stuck out was the fact that the class holds both a `DbConnection` and a `ConnectionAndTransaction` member variable so can potentially hold two different transactions at once, relying on the correctness of code in various other places to deal with this correctly. This kind of code is IME, a common source of bugs.

While I agree with the sentiment of minimal-impact, and in the long term we want to move away from nevermore, I don't think that's going to come quite as soon as we might like. Minimising hacks in the nevermore codebase I think will still hold value for some time.

If it's not too much work, personally I'd like to see this cleaned up a little, but it's not a deal breaker. Defer to @emilol for a second opinion here.

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
rosslovas: I know some people don't like asserting on exception messages but I'm ok with it, I don't envision these changing and causing friction often if ever, so I'll add this 👍

borland: To be clear: Asserting on exception messages is not good, it's brittle and annoying because messages change over time, god forbid we ever translate octopus to a non-english language.
But asserting on something generic like `InvalidOperationException` is also not good because there can be many different things that throw it, and you don't know if you've caught the right one.

My personal preference is either:
- Throw a specific exception that you can catch without ambiguity
- Throw an exception which carries some sort of programmatic error indicator, like a numeric error code, and assert on that. COMException does this with HRESULTs and I'm pretty sure the SQL/Database exceptions carry numeric error codes for similar reasons.

That said, pick your poison and do whatever is most practical here. If you're confident that transaction commit won't throw `InvalidOperationException` from any paths other than the one you expect, that's good enough.

source/Nevermore/Advanced/ReadTransaction.cs
emilol: > In this PR I was pretty torn between architecting things nicely vs a minimal change approach just to serve our needs.

There's definitely a healthy tension there. Two goals that minimising changes can serve us in a public library that others depend on like this one can be: reducing the chance of a breaking change (not relevant here, we already have a breaking change) else, minimising what needs to change when upgrading. I don't think we get the latter either in this case? 

Minimising changes also comes with the risk of introducing bugs, but as we've already said in this thread, there's also a risk of bugs in this minimal approach. 

I'd be in favour of tidying this one up a bit. I haven't had a chance to look through the changes in detail, but I do wonder if there needs to be a nevermore-specific representation of a dbConnection and/or transaction, other than what we already have. Maybe this would be worth a mob after standup?

rosslovas: No worries, I don't think there's any need to discuss further if it's worth the investment; I'll have another pass over this and go back to the approach I had originally that eliminates the redundant connection/transaction members.

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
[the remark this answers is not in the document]
rosslovas: I've removed `IConnectionAndTransaction` now so it's no longer necessary, I'm asserting on the behaviour instead now :)

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,138 @@
...
+        var nonOwnedTransaction = Store.CreateWriteTransactionFromExistingConnectionAndTransaction(sqlConnection, sqlTransaction);
+        nonOwnedTransaction.Dispose();
+
+        using (new AssertionScope())
+        {
+            // ReSharper disable once AccessToDisposedClosure
+            ((Action)(() => sqlTransaction.Commit())).Should().NotThrow();
+            sqlConnection.State.Should().Be(ConnectionState.Open);
+        }
+    }
```
emilol, line 137: Some good edge cases covered 👍 

Would it be possible to write a test case for the happy path for an externally owned connection? 

e.g. start a non-owned transaction, do some work, commit, verify changes?

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -23,20 +23,40 @@ public class WriteTransaction : ReadTransaction, IRelationalTransaction
...
         public WriteTransaction(
             IRelationalStore store,
-            RelationalTransactionRegistry registry,
+            IRelationalTransactionRegistry registry,
             RetriableOperation operationsToRetry,
             IRelationalStoreConfiguration configuration,
             IKeyAllocator keyAllocator,
-            string name = null
-        ) : base(store, registry, operationsToRetry, configuration, name)
+            Action<string>? customCommandTrace = null,
```
emilol, line 33: is this unique to each ctor invocation, or could it be on the config object?

source/Nevermore/Advanced/ReadTransaction.cs
emilol: An alternative could be to make it the default implementation. That might get tricky to get parity with the current locking behaviour though. It looks like we hold a collection of traces for the lifetime of this transaction and also use it as a locking mechanism?

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,138 @@
...
+        using (new AssertionScope())
+        {
+            const string expectedMessage = "An existing connection and transaction were provided, they should have been opened externally";
+
+            await ((Func<Task>)(async () => await readTransaction.OpenAsync(cancellationToken)))
+                .Should()
+                .ThrowExactlyAsync<InvalidOperationException>()
+                .WithMessage(expectedMessage);
+
+            ((Action)(() => readTransaction.Open()))
```
borland, line 44: Trivia: Rather than casting the lambda to Action you can do `new Action(() => readTransaction.Open()).Should()...` which has fewer brackets

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
```diff
@@ -0,0 +1,138 @@
...
+    [Test]
+    public async Task OpeningWriteTransaction_FromExistingConnectionAndTransaction_Throws()
+    {
+        await using var sqlConnection = new SqlConnection(ConnectionString);
+        await sqlConnection.OpenAsync(cancellationToken);
+        await using var sqlTransaction = sqlConnection.BeginTransaction();
+
+        var writeTransaction = (WriteTransaction) Store.CreateWriteTransactionFromExistingConnectionAndTransaction(sqlConnection, sqlTransaction);
+
+        using (new AssertionScope())
```
borland, line 70: I'm surprised the IDE didn't complain at you and cause you to change this into `using var scope = new AssertionScope();`

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -26,16 +26,19 @@ namespace Nevermore.Advanced
...
+        readonly IRelationalTransactionRegistry registry;
         readonly RetriableOperation operationsToRetry;
         readonly IRelationalStoreConfiguration configuration;
         readonly ITableAliasGenerator tableAliasGenerator = new TableAliasGenerator();
 
         readonly Func<string, DbConnection> connectionFactory;
+        readonly Action<string>? customCommandTrace;
         readonly string name;
 
+        protected bool TransactionIsNevermoreOwned { get; } = true;
```
borland, line 40: Trivia: This class is called `ReadTransaction`; the property name `TransactionIsNevermoreOwned` makes it unclear if we're talking about `this` being nevermore owned, or the internal SQL transaction being nevermore owned.

Maybe something like `OwnsSqlTransaction` or perhaps the inverse `IsSqlTransactionExternal` or something?

borland (approved):

source/Nevermore.IntegrationTests/Advanced/UseExistingTransactionFixture.cs
rosslovas: It only suggests that if you're already giving it an alias i.e. `using (var x = new AssertionScope())` 🙂

source/Nevermore/Advanced/WriteTransaction.cs
rosslovas: In theory you _could_ put it on the config object and for some simple use cases that might be sufficient, but in practice it's so much more convenient if you can directly associate the tracing with the specific transaction that the trace is coming from by just binding them together like this.

Another nice property from doing it this way is it's optional per Nevermore transaction; for externally managed transactions this is really useful, but if you're not adopting an existing transaction then it's less useful, so in a codebase that does a mix of both you might only want custom tracing in some cases and can just pass a `null` in the others.

source/Nevermore/Advanced/ReadTransaction.cs
rosslovas: Thanks, I like `OwnsSqlTransaction` so I've gone with that 👍

source/Nevermore/Advanced/ReadTransaction.cs
rosslovas: > It looks like we hold a collection of traces for the lifetime of this transaction and also use it as a locking mechanism?

Yep 👍 and Nevermore relies on that collection elsewhere like in `RelationalTransactionRegistry`, so turning it off when we provide a custom implementation would stop that from working. It may be solvable but I think it's best to keep it an additional thing right now.

emilol (approved):

PR #275, opened 2023-11-22
Add support for a custom table name provider

desruc:
This PR introduces the `ITableNameResolver` interface and a default implementation. This default implementation attempts to get the table name from the registered document maps. This is consistent with how they're currently retrieved in `DataModificationQueryBuilder`.

`ITableNameResolver` is currently **only** used in the `RelatedDocument`s feature.

### DIFF
--- a/source/Nevermore/Mapping/TableNameProvider.cs
+++ b/source/Nevermore/Mapping/TableNameProvider.cs
@@ -1,0 +1,16 @@
+﻿using System;
+
+namespace Nevermore.Mapping
+{
+    public class TableNameResolver : ITableNameResolver
+    {
+        readonly IDocumentMapRegistry mappings;
+
+        public TableNameResolver(IDocumentMapRegistry mappings)
+        {
+            this.mappings = mappings;
+        }
+
+        public string GetTableNameFor(Type documentType) => mappings.Resolve(documentType).TableName;
+    }
+}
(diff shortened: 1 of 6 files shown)


source/Nevermore/Mapping/TableNameProvider.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Nevermore.Mapping
+{
+    public class TableNameResolver : ITableNameResolver
```
emilol, line 5: this probably doesn't need to be exposed by default

source/Nevermore/RelationalStoreConfiguration.cs
```diff
@@ -23,11 +23,11 @@ public class RelationalStoreConfiguration : IRelationalStoreConfiguration
         readonly Lazy<string> connectionString;
 
 #nullable enable
-        public RelationalStoreConfiguration(string connectionString, IPrimaryKeyHandlerRegistry? customPrimaryKeyHandlerRegistry = null) : this(() => connectionString, customPrimaryKeyHandlerRegistry)
+        public RelationalStoreConfiguration(string connectionString, IPrimaryKeyHandlerRegistry? customPrimaryKeyHandlerRegistry = null, ITableNameResolver? tableNameResolver = null) : this(() => connectionString, customPrimaryKeyHandlerRegistry, tableNameResolver)
```
emilol, line 26: since this is not a pre-req and the prop is settable, we don't gain much from including it in the ctor. I notice we've used this pattern before, I think it might have just been missed that time.

emilol (approved):
A couple of minor suggestions but otherwise LGTM

PR #280, opened 2024-02-13
Make RelationalTransactionRegistry extensible

rosslovas:
This PR makes it possible to specify a custom `IRelationalTransactionRegistry` via `RelationalStoreConfiguration`.

This PR also makes `RelationalTransactionRegistry.WriteCurrentTransactions` virtual so that you can override Nevermore's implementation to include other useful information.

### Breaking changes

`ConnectionString` and `MaxPoolSize` are no longer exposed by `RelationalTransactionRegistry` as they're duplicates of information already available via `RelationalStoreConfiguration`.

`RelationalTransactionRegistry` now accepts the max SQL connection pool size via its constructor rather than the entire connection string, as that's all it actually needs.


YuKitsune (approved):
👍

PR #281, opened 2024-02-14
Add option to disable concurrency protection

adam-mccoy:
## Background

[sc-69597]

Nevermore has a feature which allows concurrent queries to be issued to a transaction without the need for MARS to be enabled within SQL Server. It does this by blocking the execution of queries on a semaphore which causes them to be issued to the database server one-at-a-time.

This feature can cause issues in some situations where one query is executed which depends on a previously issued query, creating a deadlock. Nevermore attempts to detect this situation and throw an exception, but cannot always do so reliably.

Currently this feature is on by default, with no option to disable it.

## Result

Add a configuration option `SupportConcurrentExecution` to turn this feature on or off as desired. The default will be on for compatibility.


borland (approved):
LGTM

PR #282, opened 2024-02-22
allows commiting of non owned transactions

jimmyp:


michaelongso:
Tagged Cloud PT, https://octopusdeploy.slack.com/archives/C01Q95KPRM4/p1709254205324879

jimmyp:
> Previously Commit() would throw an exception if the SQL transaction was externally owned.

You are totally right, I got myself confused while trying to use the [MS convention on TryXXX methods](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/exceptions-and-performance)... To not make this a breaking change I'll have to not use this convention

source/Nevermore/Advanced/WriteTransaction.cs
```diff
@@ -337,31 +337,57 @@ public async ValueTask<string> AllocateIdAsync(string tableName, string idPrefix
...
+            
+            if (!configuration.AllowSynchronousOperations)
+                throw new SynchronousOperationsDisabledException();
+
+            CommitTransactionAndRunHooks();
+        }
 
         public void Commit()
         {
-            if (Transaction is null)
```
gb-8, line 343: There's comment in the matching [Server PR](https://github.com/OctopusDeploy/OctopusDeploy/pull/23120/files#diff-0bb7b050da17b51ac605821c161b7e5457452f6be49c1fb2afb72b5deea29467R239) that says Commit will no longer throw if the transaction is externally owned.

Did you mean to delete that `if` (below), rather than this `null` check?

Also, no-longer throwing an exception where one was thrown before is still a breaking change, although possibly one we can live with, and not a breaking change for server's usage.

jimmyp: The comment in server was out of date and should have been removed.

This if that checks the transaction is not null has moved inside of `CommitTransactionAndRunHooks` to DRY things up.

> Also, no-longer throwing an exception where one was thrown before is still a breaking change, although possibly one we can live with, and not a breaking change for server's usage.

I dont believe we are no longer throwing exceptions here.... to make this a non breaking change I introduced `CommitIfOwnedAsync()` which is what UnitOfWork will call now.

gb-8: Ok, that makes sense then. Original behaviour of `Commit()` and `CommitAsync()` are now unchanged.
Looks good ✅

source/Nevermore/IWriteTransaction.cs
```diff
@@ -5,6 +5,8 @@ namespace Nevermore
 {
     public interface IWriteTransaction : IReadTransaction, IWriteQueryExecutor
     {
+        void CommitIfOwned();
+        Task CommitIfOwnedAsync(CancellationToken cancellationToken = default);
```
adam-mccoy, line 9: Per the feedback on https://github.com/OctopusDeploy/OctopusDeploy/pull/23120, we'd rather not introduce a different way to commit a Nevermore transaction. Rather, the consumer (Octopus Server) should be aware of whether the SQL transaction is externally owned and simply not call `Commit()` on the Nevermore transaction in that case. Happy to assist with ways to solve that.

adam-mccoy (changes requested):
Blocking this one for further discussion.

PR #284, opened 2024-06-04
Ensure the SQL connection is always closed when disposing

YuKitsune:
[sc-80865]

Updates `ReadTransaction.Dispose()` to ensure that the SQL connection is disposed, even if the other dispose methods throw an exception.

Addressed https://github.com/OctopusDeploy/Issues/issues/8841

### DIFF
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Diagnostics.Contracts;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Runtime.ExceptionServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -905,25 +906,36 @@ namespace Nevermore.Advanced
 
         public void Dispose()
         {
-            try
-            {
-                // ReSharper disable ConstantConditionalAccessQualifier
-                if (OwnsSqlTransaction)
-                {
-                    Transaction?.Dispose();
-                }
+            var exceptions = new List<ExceptionDispatchInfo>();
 
-                TransactionTimer?.Dispose();
-                DeadlockAwareLock?.Dispose();
+            if (OwnsSqlTransaction)
+            {
+                TryDispose(Transaction);
             }
-            finally
-            {
-                // ReSharper restore ConstantConditionalAccessQualifier
-                registry.Remove(this);
+            
+            TryDispose(TransactionTimer);
+            TryDispose(DeadlockAwareLock);
 
-                if (OwnsSqlTransaction)
+            if (OwnsSqlTransaction)
+            {
+                TryDispose(connection);
+            }
+
+            registry.Remove(this);
+
+            if (exceptions.Count == 0) return;
+            if (exceptions.Count == 1) exceptions.Single().Throw();
+            throw new AggregateException(exceptions.Select(x => x.SourceException));
+
+            void TryDispose(IDisposable? disposable)
+            {
+                try
                 {
-                    connection?.Dispose();
+                    disposable?.Dispose();
+                }
+                catch (Exception ex)
+                {
+                    exceptions.Add(ExceptionDispatchInfo.Capture(ex));
                 }
             }
         }
--- a/source/Nevermore/Advanced/ReadTransaction.cs
+++ b/source/Nevermore/Advanced/ReadTransaction.cs
@@ -905,22 +905,27 @@ namespace Nevermore.Advanced
 
         public void Dispose()
         {
-            // ReSharper disable ConstantConditionalAccessQualifier
-            if (OwnsSqlTransaction)
+            try
             {
-                Transaction?.Dispose();
+                // ReSharper disable ConstantConditionalAccessQualifier
+                if (OwnsSqlTransaction)
+                {
+                    Transaction?.Dispose();
+                }
+
+                TransactionTimer?.Dispose();
+                DeadlockAwareLock?.Dispose();
             }
-
-            TransactionTimer?.Dispose();
-            DeadlockAwareLock?.Dispose();
-
-            if (OwnsSqlTransaction)
+            finally
             {
-                connection?.Dispose();
-            }
+                if (OwnsSqlTransaction)
+                {
+                    connection?.Dispose();
+                }
 
-            // ReSharper restore ConstantConditionalAccessQualifier
-            registry.Remove(this);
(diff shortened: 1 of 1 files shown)


source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -905,22 +905,27 @@ DbTransaction BeginTransaction(IsolationLevel isolationLevel, string sqlServerTr
...
 
-            if (OwnsSqlTransaction)
-            {
-                connection?.Dispose();
+                TransactionTimer?.Dispose();
+                DeadlockAwareLock?.Dispose();
             }
+            finally
+            {
+                if (OwnsSqlTransaction)
```
borland, line 921: My initial thought was: Is it safe to move this until after the TransactionTimer and DeadlockAwareLock are disposed?

I did a small dig

- TransactionTimer ends up just writing to the log. That's safe wherever you put it
   - It may now be slightly less accurate; if the `connection.Dispose` takes a non-zero amount of time, that used to be captured by the Timer and now won't be. I don't think this is worth worrying about.
   
- `DeadlockAwareLock` is just a `SemaphoreSlim`, also safe to dispose in either the old or new positions.

I'm happy there, approving the PR

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -905,22 +905,27 @@ DbTransaction BeginTransaction(IsolationLevel isolationLevel, string sqlServerTr
...
+                {
+                    connection?.Dispose();
+                }
 
-            // ReSharper restore ConstantConditionalAccessQualifier
-            registry.Remove(this);
+                // ReSharper restore ConstantConditionalAccessQualifier
+                registry.Remove(this);
```
borland, line 927: If Dispose throws, the transaction will remain in the registry, and will probably be a memory leak.
We should probably put some safety code around this, or perhaps remove it before calling other code that might throw.

I'd be interested in some other opinions here though, particularly @rosslovas and @adam-mccoy who know this code better than I. 

I "soft-approve" the PR, but will wait on final approval for a second opinion

YuKitsune: I thought had a comment here but it seems GitHub lost it. I was going to ask how we should deal with this.
It's essentially a HashSet.Remove(), and I'm not concerned about it throwing. Maybe we should move this to be first in the `finally` block?

borland: Yeah I was thinking that too. Let's see what Ross or Adam thinks

rosslovas (commented):
💭 It's not an especially pretty thing, but I wonder if this would be a good opportunity to make use of `AggregateException`? The general idea here seems to be "we want to dispose as much as possible and only _then_ throw, if we must" which potentially makes this a good fit. The nice thing about it is you don't have to think as much about things like what belongs under the try and what goes in the finally, everything's on a level playing field.

It might go something like:

```C#
public void Dispose()
{
    var exceptions = new List<ExceptionDispatchInfo>();

    try
    {
        firstThing.Dispose();
    }
    catch (Exception e)
    {
        exceptions.Add(ExceptionDispatchInfo.Capture(e));
    }

    try
    {
        secondThing.Dispose();
    }
    catch (Exception e)
    {
        exceptions.Add(ExceptionDispatchInfo.Capture(e));
    }
    
    // ...

    // C# 11 list patterns would be great here but I don't think we can in this repo
    if (exceptions.Count is 0) return;
    if (exceptions.Count is 1) exceptions.Single().Throw();
    throw new AggregateException(exceptions.Select(e => e.SourceException));
}
```

Note the use of `ExceptionDispatchInfo` ([docs](https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions#capture-and-rethrow-exceptions-properly))

source/Nevermore/Advanced/ReadTransaction.cs
rosslovas: I left a top-level comment that might make worrying about this unnecessary, but if you don't go that direction then I agree; it's hard to imagine this throwing so putting it first sounds fine 🙂

YuKitsune:
@rosslovas I was thinking of implementing something _very very similar_ to this, but I wasn't sure how much effort we wanted to put into this. I have a similar thing implemented locally for `OctopusServer`.

adam-mccoy:
I'm of the opinion that `Dispose()` shouldn't throw - certainly not explicitly. What is the exception that's being thrown which we're swallowing here?

YuKitsune:
@adam-mccoy 

https://github.com/OctopusDeploy/Issues/issues/8840
https://github.com/OctopusDeploy/Issues/issues/8841

The scenario here is that we're getting a MARS error when trying to dispose the transaction:

```
System.InvalidOperationException: The connection does not support MultipleActiveResultSets.
   at Microsoft.Data.SqlClient.SqlInternalTransaction.Rollback()
   at Octopus.Core.RelationalStorage.RawRelationalTransaction.Dispose() in ./source/Octopus.Core/RelationalStorage/RawRelationalTransaction.cs:line 153
   at Autofac.Core.Disposer.DisposeAsync(Boolean disposing)
   at Autofac.Core.Lifetime.LifetimeScope.DisposeAsync(Boolean disposing)
   at Octopus.Core.Infrastructure.MessageBus.PipelineFactory.BuildAndRun[TEvent](Int64 sequenceNumber, TEvent e, IDictionary`2 metadata, String consumerGroupId, Type handlerType, CancellationToken cancellationToken) in ./source/Octopus.Core/Infrastructure/MessageBus/PipelineFactory.cs:line 53
   at Octopus.Core.Infrastructure.MessageBus.MessagePump.Dispatch[TEvent](Int64 sequenceNumber, TEvent e, IDictionary`2 metadata, CancellationToken cancellationToken) in ./source/Octopus.Core/Infrastructure/MessageBus/MessagePump.cs:line 101
```

borland:
> I'm of the opinion that Dispose() shouldn't throw

In general I agree. In this particular instance though it would swallow that MARS error, which is a legit bug; I wouldn't want to hide that. Interestingly, things like SqlClient and other I/O code tend to call their methods "Close" rather than "Dispose" - perhaps they're thinking about that same principle

adam-mccoy (approved):
I generally don't think it's a good idea to make changes that compensate for bugs introduced elsewhere. This change isn't addressing a bug in Nevermore, but rather mitigating the impact of its misuse. That said, I think it's a worthwhile change. Exceptions raised are still propagated, but ensuring the SQL connection is closed will lessen the impact of subtle bugs.

PR #285, opened 2024-06-17
Move `OPTIONS` to end in `ToListWithCountAsyncCte`

rhysparry:
When including query hints these cannot be part of a nested query as produced by `ToListWithCountAsyncCte`. This change moves any options set on the query to the end of the generated query.

Previously, the syntax generated was invalid.

An existing unit test for `ToListWithCountAsyncCte` has been updated to include a query hint.


desruc (approved):
The changes look good. It seems like this was a bug just waiting to happen. Thanks for updating the test case!

PR #288, opened 2024-07-02
Concurrency Warning

YuKitsune:
- Removes `DeadlockAwareLock` (Most calls to it were using `Lock()` and `LockAsync()` which bypassed the deadlock checks)
- Replaces `DeadlockAwareLock` with `ITransactionConcurrencyHandler`
  - Includes various implementations for different scenarios (No locking, locking enabled, locking with logging, and logging only)
  - Logging variants will log a warning when attempting to run queries in parallel on the same transaction
- Replaces `SupportConcurrentExecution` option with `ConcurrencyMode` option
- Replaces `ThreadSafeEnumerable`s with `EnumerableWithConcurrencyHandling` for clarity


source/Nevermore/Advanced/Concurrency/ITransactionConcurrencyHandler.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Threading.Tasks;
+
+namespace Nevermore.Advanced.Concurrency
+{
+    public interface ITransactionConcurrencyHandler : IDisposable
+    {
+        IDisposable Lock();
+        
+        Task<IDisposable> LockAsync(CancellationToken cancellationToken);
+    }
```
YuKitsune, line 12: We decided to replace the `DeadlockAwareLock` with this abstraction as a way for us to enable logging without locking. It's a bit much, but felt cleaner than introducing a bunch of additional checks into a single class.
Open to alternatives.

YuKitsune: Here's a spike of what this would look like if we had a single type: https://github.com/OctopusDeploy/Nevermore/pull/289

source/Nevermore/Advanced/Concurrency/NoOpConcurrencyHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class NoOpConcurrencyHandler : ITransactionConcurrencyHandler
+    {
+        public IDisposable Lock()
+        {
+            return NoopDisposable.Instance;
+        }
+
+        public async Task<IDisposable> LockAsync(CancellationToken cancellationToken)
+        {
+            await Task.Yield();
```
borland, line 17: AFAIK Task.Yield incurs a thread-jump. Given we aren't trying to use it to capture Task.Current anymore, looks like we could replace with `return Task.FromResult(NoopDisposable.Instance)`. Nevermore doesn't have the weird "must async everything" convention that Server does.

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -521,8 +531,9 @@ IEnumerable<TRecord> Execute()
                     yield return item;
             }
 
-            return configuration.SupportConcurrentExecution
-                ? new ThreadSafeEnumerable<TRecord>(Execute, DeadlockAwareLock)
+            // Use the thread safe variant if we know it's going to be used
+            return configuration.ConcurrencyMode is not ConcurrencyMode.NoLock
```
borland, line 535: Suggestion: A clearer and more reliable way to make this decision could be

- Add a `bool IsThreadSafe { get; }` property to `TransactionConcurrencyHandler`
- The implementations can return true/false appropriately

`return TransactionConcurrencyHandler.IsThreadSafe ? new ThreadSafeEnumerable<TRecord>...`

borland: Underlying principle here: You've got one input - the `ConcurrencyMode` enum, but you make decisions about it in two places. If the decisions don't relate to each other, that's fine -- however in this case the choice to use ThreadSafeEnumerable, and the behaviour of the `TransactionConcurrencyHandler` are _highly_ related.

Technique: Reduce the decision-points (and thus, risk of bugs) by having one decision "flow through" to the next. If that all sounds like gibberish, hit me up for a zoom tomorrow :smile:

source/Nevermore/Advanced/ReadTransaction.cs
```diff
@@ -651,17 +663,17 @@ public Task<int> ExecuteNonQueryAsync(string query, CommandParameterValues? args
 
         public int ExecuteNonQuery(PreparedCommand preparedCommand)
         {
-            using var mutex = configuration.SupportConcurrentExecution 
-                ? DeadlockAwareLock.Lock()
+            using var mutex = configuration.ConcurrencyMode is not ConcurrencyMode.NoLock
```
borland, line 666: Our no-lock ConcurrencyHandlers already stub out the Lock function, so why do we need this check at all?
Can we just `using var mutex = TransactionConcurrencyHandler.Lock()` and be done with it?

source/Nevermore/ConcurrencyMode.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Nevermore
+{
+    public enum ConcurrencyMode
```
borland, line 3: Food for thought:

```csharp
[Flags]
public enum ConcurrencyMode
{
    Lock = 1
    Log = 1 << 1
}
```

This models the problem better (locking and logging are two different dimensions, and can both move independently), but is probably a worse fit for your code, as you have four implementations of the handler.

borland (commented):
I'm quite happy with the removal of DeadlockAwareLock, the choice of four implementations of concurrency handler, and the interface is nice. I would appreciate a quick zoom though, I'd like someone to show me all the places you found where the deadlockawarelock wasn't working properly.

Some comments about the implementation details in ReadTransaction, nothing obviously bad, but it'd be nice to have a chat about them before I approve.
Note: I'm not blocking; if someone else is happy to approve before I get there and you want to merge, fair game

source/Nevermore/Advanced/ReadTransaction.cs
YuKitsune: I'm keen to follow up here on zoom. Another concern here is that we still want to use the logging-only concurrency handler.
I think the cleanest way forward is to always return the `ThreadSafeEnumerable`, and give it what ever concurrency handler we've got. The only thing I don't like about that is the fact that we can return a `ThreadSafeEnumerable` with a concurrency handler that doesn't provide thread safety, but we need it for the logging anyway 😕

source/Nevermore/Advanced/Concurrency/LockOnlyConcurrencyHandler.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Nito.AsyncEx;
+
+namespace Nevermore.Advanced.Concurrency
+{
+    public class LockOnlyConcurrencyHandler : ITransactionConcurrencyHandler
```
emilol, line 8: Given this is a public library and this an internal implementation detail a sensible default here would be `private`, otherwise this will default to forming part of the public contract.

YuKitsune: Good call, fixed!

borland (approved):
All looking good, approved!

PR #290, opened 2025-01-28
Pin FluentAssertions to v7

YuKitsune:
[sc-101162]

FluentAssertions v8 changes to a non-open-source license; Commercial Use requires a paid subscription at $129 USD per developer per year.

This PR pins the version of FluentAssertions to 7.1.0.


borland (approved):

YuKitsune:
Closing this as Dependabot, Renovate and Rider do not respect NuGet version ranges in `csproj` files.
We should instead configure our dependency upgrader (Dependabot/Renovate) to prevent upgrades beyond v7.

